PR #8, opened 2018-12-03
Add protvista uniprot buttons

ljgarcia:


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "root",
   "private": true,
   "scripts": {
-    "bootstrap": "lerna bootstrap --hoist",
+    "bootstrap": "lerna bootstrap --concurrency=1 --mutex network --hoist",
```
xwatkins: What are these extra options for?

ljgarcia: Oh! sorry, that should not be there. I had to add them for a windows specific thing that was solved with admin permissions at the end, will remove them and commit again

PR #10, opened 2019-03-05
Feat/filter ui

vivekimsit:
Changelog:

- Create `protvista-checkbox` component.
- Create `protvista-filter` component.
- Wire filter component with variation adapter via `protvista-manager`


packages/protvista-variation-adapter/src/protvista-variation-adapter.js
```diff
@@ -1,11 +1,54 @@
...
+      const { sequence, variants } = this._adaptedData;
+      newValue = newValue.trim();
+      if (!newValue) {
+        this._fireEvent('load', { payload: {sequence, variants: []} });
+        return;
+      }
+      const filteredVariants = newValue
+        .split(',')
+        .map(name => getFilter(name))
+        .map(f => f.applyFilter(variants))
```
xwatkins: The issue here is that this provides the union of all filters and therefore results in some overlaps (duplications in the resulting data). 
For instance when it first loads (no filters enabled) it returns 385 items, but if I click on "Likely disease" the number jumps up to 770.

packages/protvista-variation-adapter/src/protvista-variation-adapter.js
```diff
@@ -1,11 +1,54 @@
...
+        .flat();
```
xwatkins: I'm not 100% sure whether this requires another polyfill to be imported for IE/Edge...if so it would be great to find an alternative way.

packages/protvista-filter/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "protvista-filter",
+  "version": "1.0.0",
```
xwatkins: The version will be managed by lerna - not sure how it works for the first release of a component (haven't tried yet) but maybe take the version from `lerna.json` for now.

packages/protvista-filter/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "Filter component for variants",
+  "files": [
+    "src"
```
xwatkins: also add `dist` please (for the CDN)

packages/protvista-variation-adapter/src/protvista-variation-adapter.js
vivekimsit: Done (using lodash version)

packages/protvista-filter/package.json
vivekimsit: Done, Thanks for pointing out.

packages/protvista-filter/package.json
vivekimsit: Done

packages/protvista-variation-adapter/src/protvista-variation-adapter.js
vivekimsit: Done, I am taking `accession` as key for uniqueness, please check if that is correct.

Issue #22, opened 2019-04-25
Track bumping algorithm issues


xwatkins:
There seems to be a few issues with the non-overlapping layout:
- Instead of being centered, the features appear at the bottom
- the heigh of features doesn't seem to be equal

PR #32, opened 2019-06-06
Bug/track bumping issue22

xwatkins:
Refactored track bumping in order to solve Issue #22 
Quick visual check with InterPro track seems ok, but please make sure before merging

### DIFF
--- a/packages/protvista-track/src/__tests__/NonOverlappingLayout.spec.js
+++ b/packages/protvista-track/src/__tests__/NonOverlappingLayout.spec.js
@@ -45,7 +45,7 @@ describe("Layout bumping", () => {
     expect(layout.getFeatureYPos(features[2])).toEqual(2);
   });
 
-  test("should be back on first row", () => {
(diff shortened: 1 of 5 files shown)


packages/protvista-track/src/__tests__/NonOverlappingLayout.spec.js
```diff
@@ -0,0 +1,55 @@
...
+
+  test("should be on second row", () => {
+    expect(layout.getFeatureYPos(features[1])).toEqual(8);
+  });
+
+  test("should be back on first row", () => {
+    expect(layout.getFeatureYPos(features[2])).toEqual(2);
+  });
+
+  test("should be back on first row", () => {
```
gustavo-salazar:
```suggestion
  test("should be back on third row", () => {
```

gustavo-salazar:
Cool.  Ilike the simplified version of the NonOverlappingLayout. There is only one comment in the tests but otherwise happy to merge

PR #34, opened 2019-06-13
Secondary structure supported in ProtVista Track

Swaathik:


gustavo-salazar:
I haven't check the details of the PR, but the first thing I notice is all the new `yarn.lock` files. 
@xwatkins do you know if we didn't have those files before because we weren't using _yarn workspaces_? Or is it because @Swaathik installed _nightingale_ from scratch?

Is it something we should keep at that level(per package)? or is it enough to have only one `yarn.lock` in the root of the repo.

packages/protvista-track/src/FeatureShape.js
```diff
@@ -4,233 +4,620 @@
...
+      "L" +
+      this._ftWidth +
+      ",0" +
+      this._getBrokenEnd() +
+      "L0," +
+      this._ftHeight +
+      "Z"
+    );
+  }
+  _helix() {
```
gustavo-salazar: Unfortunately prettier has made it hard to identify that the addition to the file is the last 2 shapes.

The created path is also harder to read, maybe we should rewrite this file using string literals \`\`??

PR #35, opened 2019-06-21
'dist' for CDN and 'module' for module export

xwatkins:


packages/interaction-viewer/package.json
```diff
@@ -11,6 +11,10 @@
     "uniprot",
     "ebi"
   ],
+  "files": [
+    "src",
+    "dist"
```
gustavo-salazar: I just read the files part in https://docs.npmjs.com/files/package.json 

> Files included with the “package.json#files” field cannot be excluded through .npmignore or .gitignore.

but we have dist in the git ignore, is that going to be an issue?

xwatkins: We've had this in most modules for a while now, I guess what happens is that at the time of publishing they get included regardless. According to this we should remove `/dist` from `.gitignore`?

gustavo-salazar:
So are this changes making the index.js redundant in all the packages?
I'm still confuse, in where the web component is getting loaded? especially if someone is using the components directly from CDN

xwatkins:
No, `index.js` is only redundant when loading modules. For the CDN, if you look in the `webpack` build file, `index.js` is the entry point.

gustavo-salazar:
@xwatkins  what are the `/temp` files? should they be part of the commit?

xwatkins:
Woops no, removed!

packages/interaction-viewer/package.json
gustavo-salazar: I think the current effect that we have is that `/dist` gets included in npm but is not included in github, which I think is what we want.
So I guess, just leave it there.

gustavo-salazar:
OK got it! 
The entry point  for webpack is `index.js`, and therefore webpack creates the `/dist` file already including the loading of the web components, so it will work from CDN!
But whoever uses the package as a dependency it will could use the component in `/src` and do the load of the web component separately.

PR #37, opened 2019-07-01
Components register with manager instead of other way round

xwatkins:
The current implementation relies on components to be present in the DOM when `protvista-manager` is connected to the DOM. This means that components (such as tracks) adding/removed dynamically can't be controlled by `protvista-manager`.
This implementation uses a "bottom-up" approach, allowing the components to register with `protvista-manager` instead.

- Added a `register(element)` function to `protvista-track`
- Removed the `isManaged` flag, and added a call to register the components in the `connectedCallback()`
- Added a `_polyfillElementClosest()` function to polyfill the `Element.closest` function (https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)
- The change listener is now added only once, directly to `protvista-manager`


aurel-l:
If you had a MutationObserver on the manager before, shouldn't it be where it would watch for changes to the subtree of the manager and tracks and other components would then be added to the manager, even if added dynamically?

xwatkins:
Yes, I think this approach (with `MutationObserver`) was working fine initially, but as the number of components grew we ended up having to add an `isManaged` flag to components which needed to be updated by `protvista-manager`. This kind of defeats the purpose of having the top-down approach as the overhead of having code in the components to register themselves with `protvista-manager`. There has also been some issues in the past about `MutationObserver` not working correctly with the Shadow DOM, which some components use.

PR #45, opened 2019-07-26
Protvista saver

Swaathik:


gustavo-salazar:
Maybe expose the file name and extension as attributes of the component. And check if is there any other setting from rasterizeHTML that can be xposed.

Swaathik:
File names and the formats have been included as attributes. Also the image dimensions are changed according to the window size rather than fixed values

PR #59, opened 2019-11-27
Use shift to zoom

gustavo-salazar:
# Description of changes
Adding the attribute `use-shift-to-zoom` to zoomable. When present the zoom events are only considered, while shift is pressed.

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide => Not applicable


aurel-l:
I like this behaviour, but since we're going to have this behaviour available, shouldn't we put that as default?
I like the way embedded Google Maps do it. That is, don't intercept scroll for zoom (let it scroll the page normally) but also displays an overlay saying that shift+scroll to zoom is available.

gustavo-salazar:
> I like this behaviour, but since we're going to have this behaviour available, shouldn't we put that as default?
I put it as an opt-in jus t o kept what it was before, and that the change won take by surprise to any user. So whoever wants to use [shift] to zoom, has to explicitly say so.

> I like the way embedded Google Maps do it. That is, don't intercept scroll for zoom (let it scroll the page normally) but also display an overlay saying that shift+scroll to zoom is available.

I also like that, and it is something that, I'm considering for the interpro website, but I think that whould be a decision of the developer using the web component. But i think we can use that approach for the*magic* one!

xwatkins:
I agree that this should be the default behaviour. In the new UniProt website we are using Nightingale component throughout the entry page (https://ebi-uniprot.github.io/uniprot-website/uniprotkb/P05067) and the scrolling has been a big issue.

I like how Google maps have defined `gestureHandling` and we could do something similar (default to auto based on whether the window is scrollable or not): https://developers.google.com/maps/documentation/javascript/interaction#gestureHandling

In regards to the overlay, I wonder if this would sit better in the manager component as it is the component which enables the zooming functionality? My issue with leaving it up to the resource or having it in `protvista-uniprot` is that of consistency in the way it looks.

One last thing: Google uses Ctrl or ⌘ to enable zooming, while LiteMol uses Ctrl+click so I wonder if Ctrl would be more generic?

aurel-l:
So about the changing behaviour we should keep it opt-in for now (within a minor release), and change it in the next major version.

Also, yeah, the "display something to tell the user to ctrl-scroll" should be the responsibility of the ✨️magic✨️ components, or whoever uses the components directly. So as long as we let the scroll event bubble up we should be good to go.

gustavo-salazar:
I decided to go for [SHIFT] instead of [CTRL] out of laziness, just to avoid having to consider the case that in Mac is more common to use ⌘ instead of [CTRL]

xwatkins:
> So about the changing behaviour we should keep it opt-in for now (within a minor release), and change it in the next major version.

Sounds good. So do we agree to go for `Ctrl`? What do you think of the idea of replicating the definition in the Google Maps API (meaning the new option would be `gesture-handling: cooperative` instead of `use-shift-to-zoom`)

Note: We should probably make a list of what we'd like the next major version to have. For instance bringing the `@nightingale` scope in the npm package definition.

>Also, yeah, the "display something to tell the user to ctrl-scroll" should be the responsibility of the ✨️magic✨️ components, or whoever uses the components directly. So as long as we let the scroll event bubble up we should be good to go.

...but in our case we want the same overlay look and feel for the ✨component and single tracks with a navigation and manager component... maybe the solution is to create a `protvista-overlay` component, used by `protvista-uniprot` and available for anyone to use?

gustavo-salazar:
I like the `protvista-overlay` idea... I'll work on it!

About using `[CTRL]` I'm still a bit hesitant, because Mac OS prefers `[⌘]` over `[CTRL]` and not sure is worth adding logic just for that... or we can just use `[CTRL]` no matter if is Mac or not.

gustavo-salazar:
@aurel-l and @xwatkins 
* I created the overlay component and added to the manager example in the App.
* I also change the code so it uses `[CTRL]` instead of `[SHIFT]`.
* In the process I created a Hello world tutorial

* Tested on latest stable versions of Chrome, Firefox and Safari.
Please review!

xwatkins:
That's great!! Only one comment, how about centering the text vertically with:
```
display: flex;
align-items: center;
justify-content: center;
```
on the div?

packages/protvista-overlay/src/ProtvistaOverlay.js
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  render() {
+    let html = "";
+    const target = document.getElementById(this.for);
+    if (target) {
+      const { height, width } = target.getBoundingClientRect();
+      html = `
+      <div style="
+        display: block;
```
xwatkins: how about centering the text vertically with:
```
display: flex;
align-items: center;
justify-content: center;
```

gustavo-salazar:
@xavier Cool, message centered!

PR #62, opened 2019-12-10
Structure improvements

its-mahdi:
### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide


packages/protvista-structure/README.md
```diff
@@ -1,13 +1,41 @@
 # `<protvista-structure>`
+
 In progress - A native web component wrapper around LiteMol
 
 ## Usage
+
 ```html
-      <protvista-structure id="1AAP"></protvista-structure>
+<!-- Basic usage -->
+<protvista-structure id="1AAP" />
```
xwatkins: I don't think this would work on its own as it relies on the accession. Also `id` is not mentioned, I think that was an error in the doc.

packages/protvista-structure/README.md
```diff
@@ -1,13 +1,41 @@
...
 
 ### Properties
+
 #### `accession: UniProt accession`
-A protein accession
\ No newline at end of file
+
+A protein accession
+
+#### `highlightresidues`
```
xwatkins: only `highlight` is valid, not `highlightresidues` (which needs to be removed). 
The description would be something like "highlight residues corresponding to the given sequence positions".

packages/protvista-structure/README.md
xwatkins: Actually I think it would be nice if we could bypass the `loadUniProtEntry()` if `accession` is not passed (which would also mean no table would be displayed).

packages/protvista-structure/src/protvista-structure.js
```diff
@@ -129,7 +135,11 @@ class ProtvistaStructure extends HTMLElement {
...
     this.appendChild(flexContainer);
     flexContainer.appendChild(litemolDiv);
-    flexContainer.appendChild(this.tableDiv);
+    if (!this.hideTable) {
+      this.tableDiv = document.createElement("div");
+      this.tableDiv.className = "table-container";
+      flexContainer.appendChild(this.tableDiv);
+    }
     this.loadLiteMol();
     this.loadUniProtEntry().then(entry => {
```
xwatkins: Can we bypass this if we don't show the table? It adds an unnecessary call to the UniProt API.

packages/protvista-structure/README.md
its-mahdi: For the record: We won't be able to load the structure from PDBe call directly without knowing the protein accession we are interested in, because the PDBe call returns the structure details one level below the protein accession.

packages/protvista-structure/src/protvista-structure.js
```diff
@@ -103,58 +112,87 @@ class ProtvistaStructure extends HTMLElement {
...
+
+  get molecule() {
+    return this.getAttribute("molecule");
+  }
+
+  get height() {
+    return this.getAttribute("height");
+  }
+
+  get hideViewportControls() {
```
xwatkins: I don't think this is used?

packages/protvista-structure/src/protvista-structure.js
```diff
@@ -103,58 +112,87 @@ class ProtvistaStructure extends HTMLElement {
...
-        this.style.display = "none";
-        return;
-      }
-      this.loadStructureTable();
-      this.selectMolecule(
-        this._pdbEntries.filter(d => d.properties.method !== "Model")[0].id
-      );
-    });
+
+    if (!this.hideTable || !this.molecule) {
```
xwatkins: There is a case where we both the table to be loaded from UniProt AND `molecule` to be specified. I don't think this case is being taken care of?

packages/protvista-structure/src/protvista-structure.js
its-mahdi: No, it is used in line 16:
    this._height = this.height || "480px";

`this.height` is the getter where line 124 is the body of the function.

packages/protvista-structure/src/protvista-structure.js
its-mahdi: I'm not sure if I follow. This `if` says if we are not hiding the table or we have passed a molecule, then load the required data from UniProt. Wouldn't that cover the use-case since the molecule is provided?

PR #65, opened 2020-01-16
Highlight

xwatkins:
### Reference to issue
The requirement was to be able to establish a direct link between features in tracks and features in the data table component, scrolling into position and highlighting the corresponding feature when selected in a track.

### Description of changes
- As not all features have a unique identifier, I decided to add `protvistaFeatureId` to each feature in the data adapters at the time the data is transformed (note: didn't do the InterPro adapter). Using the `uuid` library for simplicity.
- Because the ids were then generated at each transformation, I needed a way to share the transformed data between components (something we started working towards during the biohackathon). I added a _subscribers_ attribute to `protvista-feature-adapter` which takes a list of html selectors. This provides another way of using data transformers and is now documented.
- Changes to `protvista-datatable` to implement the scrolling and highlight. Added a `selectedid` attribute.

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide


gustavo-salazar:
So the `subscribers` are implemented in all the `adapters`? 
Shouldn't we then create an adapter that has that implementation of it and all the adapters inherit from it?

gustavo-salazar:
Ups...thats actually what you are doing and the other adapters inherit from `ProteinFeatureAdapter` 🤦🏽‍♂️

xwatkins:
Thanks!

PR #77, opened 2020-04-23
Added a basic project example using vue

sacdallago:
### Reference to issue

I can find no issue about this

### Description of changes

I've created a folder `examples` in the root. In that folder, I created a vue project like so `vue create basic_vue ` and keeping all default options. I've then installed a few `protvista-*` components and `d3`, and changed the default "dummy"/HelloWorld component in Vue to display a very basic visualization.

I think this is important as it gives a blueprint for people interested in using protvista components in a vue app. It's extremely minimal (and purposefully kept so) as the key highlights are:
 - how to register components when in Vue
 - how to ignore components through `ignoredElements`
 - how to bind data in a vue component

### Tests / Styleguide
 - [ ] Tests pass
 - [ ] Styleguide


xwatkins:
Great idea! Will add the same for React.

Would you be able to link to the repo and add a bit of documentation [here](https://github.com/ebi-webcomponents/nightingale/blob/master/documentation/GETTING_STARTED.md#usage-with-popular-libraries-and-frameworks)?

sacdallago:
Will add to the docu over the weekend :)

PR #79, opened 2020-05-29
Alignment Component

gustavo-salazar:
## Description of changes
I added a new component to display multiple sequence alignments.
It uses the project in https://github.com/ebi-webcomponents/react-msa-viewer which is forked of a project that rewrites the BIoJS MSA component using react.
This is a very first prototype, and it doesn't include many of the options of the library, but for now the purpose of it is to make sure that it can play along other nightingale components

A demo of it in codepen: https://codepen.io/gustavo-salazar/full/rNxaMgb

**Note:** Because this library used react I have to do some changes in the webpack config. Please have a close look to it, checking the new build still works for your components.

### Tests / Styleguide
 - [ x ] Tests pass
 - [ x ] Styleguide


sacdallago:
@gustavo-salazar et al. You might want to consider also https://github.com/sanderlab/AlignmentViewer2.0 instead of Sebastian's and Rob's MSAviewer. It starts from the same ideas and principles as MSAviewer but @npgauth is aiming at making it reactive and fast for extremely deep alignments

gustavo-salazar:
Hey @sacdallago 
Thanks for the heads up. This pull request is a first proof of concept with the `react-msa-viewer`. Before choosing to work with it I checked out a couple of other projects, unfortunately I didn't knew about AlignmentViewer2.0, otherwise I would've included in my research state.
BTW, the `react-msa-viewer` is not the original MSA that Seb and Rob put in BioJS; it is a complete rewrite using more up-to-day technologies, annd on my tests it have perform quite nicely in big alignments. But as I said, this is a proof-of-concept, once the PR is accepted, I'll tested how it behaves dealing with seed and full Pfam alignments. And depending on it I'll continue using it or look for an alternative, in which case I'd consider the library you are suggesting. Which based on its demo, looks quite promising.
Also if you are interested in web libraries to view alignments, have a look to @ihh 's one:  https://github.com/ihh/abrowse it has a very cool feature to collapse a subset of alignments into a logo view!

sacdallago:
@gustavo-salazar all good :) just wanted to make you aware.

I was under the impression that @npgauth wasn't fully done with 2.0 (it's still under development, from what I understand; don't want to talk for Nick :)), but since I saw this PR I thought it might be good to create the bridge now, rather then when it's all said and done ;)

gustavo-salazar:
In my last commit I have made another change to the webpack config:`    libraryTarget: "umd",`

In theory this shouldn't affect future deployments because `umd` exposes the library under all the module definitions, including the default `var` that we were using until now. See https://webpack.js.org/configuration/output/#expose-a-variable

Nonetheless, please let me know if you think that can cause other issues that i didn't consider.

PR #81, opened 2020-06-04
Textarea sequence

gustavo-salazar:
### Reference to issue
N/A

### Description of changes
Adding a new component:  `<textarea-sequence />`
The readme of the component describes its parameters and methods.
I published on a canary version and you can play with a demo here: https://codepen.io/gustavo-salazar/full/PoZwNwJ

**Note:** I changed a bit the configuration of typescript in webpack, from `babel-typescript` plugin to `ts-loader` webpack loader. I checked `data-loader` in the example app and seems to be OK. But please check yourselves.

### Tests / Styleguide
 - [x] Tests pass
I haven't added test to this package (I should). But existing tests are passing
 - [ ] Styleguide


gustavo-salazar:
Ok @aurel-l requested changes done so please check again. The only thing I didn't implement is the one about exposing transformations. I think that can lead to issues about expecting the raw_seuqence to be something else. And by exposing, the functions to clear, I think is enough.

PR #102, opened 2020-09-08
Variation model changes

xwatkins:
### Tests / Styleguide
 - [ ] Tests pass
 - [ ] Styleguide


packages/protvista-variation-adapter/src/ProtvistaVariationAdapter.ts
```diff
@@ -1,38 +1,43 @@
 import ProtvistaFeatureAdapter from "protvista-feature-adapter";
 import { v1 } from "uuid";
+import { NightingaleElement } from "data-loader";
```
aurel-l, line 3: There's no reason that this definition should be kept in `data-loader`, I guess it'd be good to have a common place not in any specific package with all of the common code and interface. Do you know if there's a way to do that in lerna? Or do we just import from a relative path up outside the packages themselves?

app/src/components/ProtvistaVariation.jsx
```diff
@@ -1,35 +1,38 @@
...
-            <data-loader>
-              <source src="https://www.ebi.ac.uk/proteins/api/variation/P05067" />
-            </data-loader>
-          </protvista-variation-adapter>
-        </protvista-variation>
-      </Fragment>
-    );
-  }
-}
+    window.addEventListener("change", (e) => {
```
aurel-l, line 16: You'd need to unregister this on unmount, because that might cause issues, especially if it's listening to all change event within the whole window

aurel-l (approved):

PR #106, opened 2020-09-15
Define transformed type

xwatkins:
### Tests / Styleguide
 - [ ] Tests pass
 - [ ] Styleguide

### DIFF
--- a/packages/protvista-variation-adapter/src/protvista-variation-adapter.ts
+++ b/packages/protvista-variation-adapter/src/protvista-variation-adapter.ts
@@ -3,7 +3,14 @@ import { v1 } from "uuid";
 import { NightingaleElement } from "data-loader";
 
 import formatTooltip from "./tooltipGenerators";
-import { ProteinsAPIVariation, Xref, SourceType } from "./variants";
+import {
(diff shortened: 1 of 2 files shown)


packages/protvista-variation-adapter/src/protvista-variation-adapter.ts
```diff
@@ -13,10 +20,32 @@ const getSourceType = (xrefs: Xref[], sourceType: SourceType) => {
   return xrefNames;
 };
 
-export const transformData = (data: ProteinsAPIVariation) => {
+export type TransformedVariant = {
```
aurel-l, line 23: If I got this correctly, this is the data in the shape that is required by protvista-variation, right?
Maybe this should actually go into this package and get in imported here for this function's output.

aurel-l (approved):
I know the change I suggested means making yet an other package switch to TS now. So happy to approve that for now, knowing that we will eventually have to make that change

xwatkins:
There is actually also an extra level of transformation which is currently done by `protvista-variation` and which I would like to eventually move to `protvista-variation-adapter`.

PR #107, opened 2020-09-16
Protvista track feature improvement

Swaathik:
### Reference to issue
The event is triggered on hovering over the feature shape. In cases like helix, or other shapes different than rectangle, it causes the event to trigger every time the mouse pointer moves between the path resulting in flickering like in https://github.com/ProteinsWebTeam/interpro7-client/issues/242

### Description of changes
An extra group is added `<g class="fragment-group"><path /><rect /></g>` in place of just the `<path>` and the rectangle is  responsible for event handling.

Could you please let us know if the above change is okay and goes well at your end?

### DIFF
--- a/packages/protvista-track/src/protvista-track.js
+++ b/packages/protvista-track/src/protvista-track.js
@@ -207,6 +207,39 @@ class ProtvistaTrack extends ProtvistaZoomable {
         return feature.opacity ? feature.opacity : 0.9;
       })
       .call(this.bindEvents, this);
+
+    this.rectangles = this.locations
+        .selectAll("g.fragment-group")
+        .data((d) =>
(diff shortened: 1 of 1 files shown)


aurel-l (approved):
I see, that seems like a sensible approach to have a bigger event target than just the actual shape itself 👍

gustavo-salazar:
> I see, that seems like a sensible approach to have a bigger event target than just the actual shape itself 👍
Yeah just worried that there might be a selector somewhere  that requires `path.feature` not to be embedded in `g.fragment`. I don't think there is but we just want to be sure, @xwatkins do you think this might be an issue??

If you guys happy with it we will merge and publish

PR #110, opened 2020-09-23
adding a higher order component withMargin

### LINKED ISSUE #109: Navigation bar is not considering the margin used in track.
**Describe the bug**
All tracks inherited from `zoomable` and therefore they have an attribute margin (https://github.com/ebi-webcomponents/nightingale/blob/e2d195d5596fb1867017ca3c49a26ba6a81463b2/packages/protvista-zoomable/src/protvista-zoomable.js#L147-L154).
The `navigation` component doesn't extends `zoomable` and therefore it doesn't have the attributte `margin`, so it doesn't take it into consideration when rendering.

**To Reproduce**
1. Go to https://ebi-webcomponents.github.io/nightingale/#/manager
2. Make sure the naviagation is starting on the 1st residue.
3. Notice the misalignment (10px) with other components.

**Expected behavior**
The navigation should align with other components.

**Screenshots**
Check the corner at the left site where the final point of it should align with the start of the first base in the sequence, but it doesn't.
Screenshot 2020-09-22 at 14 32 34

gustavo-salazar:
### Reference to issue
FIX #109 

### Description of changes
It creates a higher order component that embeds the logic for a margin object. it exposes the following attributes and parameters:
* `marginleft`
* `marginright`
* `margintop`
* `marginbottom`

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide

### DIFF
--- a/packages/protvista-utils/src/withMargin.js
+++ b/packages/protvista-utils/src/withMargin.js
@@ -1,0 +1,48 @@
+const sides = ["left", "right", "top", "bottom"];
+const marginSides = sides.map((side) => `margin${side}`);
+
+const withMargin = (
+  Element,
+  options = {
+    initialValue: {
+      top: 0,
+      bottom: 0,
+      left: 0,
+      right: 0,
+    },
+  }
+) => {
+  for (const side of sides) {
+    Object.defineProperty(Element.prototype, `margin${side}`, {
+      get() {
+        return this.margin[side] || 0;
+      },
+      set(value) {
+        this.margin[side] = +value;
+      },
+    });
+  }
+
+  class ElementWithMargin extends Element {
+    constructor() {
+      super();
+      this.margin = options.initialValue;
+    }
+
+    static get observedAttributes() {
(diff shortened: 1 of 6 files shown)


packages/protvista-utils/src/withMargin.js
```diff
@@ -0,0 +1,48 @@
...
+    initialValue: {
+      top: 0,
+      bottom: 0,
+      left: 0,
+      right: 0,
+    },
+  }
+) => {
+  for (const side of sides) {
+    Object.defineProperty(Element.prototype, `margin${side}`, {
```
aurel-l, line 16: Maybe this should be applied on `ElementWithMargin` and not on `Element`, as it will modify the original element's prototype unnecessarily as the one that's really going to be used will be the returned `ElementWithMargin`?

gustavo-salazar: Cool, I see there is a potential for side effects with the current code, I'll change it as you suggest.

aurel-l (approved):

PR #111, opened 2020-09-24
Protvista links

gustavo-salazar:
### Description of changes
New protvista track: `protvista-links`: It displays links in between residues as nodes for the residues with links and arcs between linked ones. The arcs are only visible on hover.

---
It also includes several changes in config files for better support for typescript+eslint and prettier.

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide


packages/protvista-links/src/links-parser.ts
```diff
@@ -0,0 +1,81 @@
...
+      sets[n2set[n1]].push(n2);
+    } else if (!n2set[n1] && n2set[n2]) {
+      n2set[n1] = n2set[n2];
+      sets[n2set[n1]].push(n1);
+    } else if (n2set[n1] && n2set[n2] && n2set[n1] !== n2set[n2]) {
+      sets[n2set[n2]].forEach((n) => {
+        sets[n2set[n1]].push(n);
+        n2set[n] = n2set[n1];
+      });
+    }
```
dlrice, line 34: I don't understand the context completely so this question might not make sense: do you ever care about the case where `n2set[n1] && n2set[n2] && n2set[n1] === n2set[n2]`?

app/src/components/ProtvistaMSA.jsx
```diff
@@ -174,68 +159,46 @@ const ProtvistaMSAWrapper = () => {
...
+              textTransform: "uppercase",
             }}
+          >
+            Contacts
+          </div>
+          <protvista-links
+            id="links-track"
+            length={sequence.length}
+            height={20}
+            use-ctrl-to-zoom
```
dlrice, line 198: Is this needed?

dlrice (commented):
A few observations when using the Alignments demo:
* Moving the range in protvista-navigator only navigates the links and not the MSA
* The MSA disappears when I mouse over a link node so wasn't able to see the highlight event

gustavo-salazar:
> A few observations when using the Alignments demo:
> 
> * Moving the range in protvista-navigator only navigates the links and not the MSA
> * The MSA disappears when I mouse over a link node so wasn't able to see the highlight event
Hey Daniel, That's weird, I have the demo running here: https://www.ebi.ac.uk/~gsalazar/nightingale/#/msa and it seems to work ok. Or at least it works on my laptop :-)

Have you updated the `react-msa-viewer`, the errors you describe sound to me as if is using a previous version of it.

packages/protvista-links/src/links-parser.ts
gustavo-salazar: So the `parseLinksAssociative` idea is that if position 20 is linked with 30, and position 30 is linked with 40, then I can infer that position 20 is linked with 40. Which I thought was the logic of the contacts I need to display, but it is not, a link has to be explicitly included. So I'm not using this function right now, but I though of leaving it there in case there is a future case for it.

in any case, this is logic to ignore the case `n2set[n1] && n2set[n2] && n2set[n1] === n2set[n2]`:
* if `n2set[n1]` exists it implies that `n1` belongs to that set
* if `n2set[n2]` exists it implies that `n2` belongs to that set
* if `n2set[n1] === n2set[n2]`, they are the same set, so I don't need to include anything in the set because they are already there, and any associative link has already taken care of.

app/src/components/ProtvistaMSA.jsx
gustavo-salazar: do you mean the `use-ctrl-to-zoom` attribute?
If so, it's not needed, it is a matter of preference. 
Zooming in the track is an inherited feature from `protvista-zoomable`, which by default uses scrolling gestures to zoom in/out. We had several complains of users saying that behaviour was disrupting because they wanted to scroll and not to zoom. so we add the `use-ctrl-to-zoom` attribute, to say that this track would use scrolling gestures for zooming only if the `[ctrl]` key is pressed.

dlrice: > Have you updated the `react-msa-viewer`, the errors you describe sound to me as if is using a previous version of it.

You're quite right. I had deleted all of the `node_modules` directories and ran `yarn` but I guess this wasn't enough so I recloned the `nightingale` and removed the `react-msa-viewer` repo.

dlrice:
> Have you updated the `react-msa-viewer`, the errors you describe sound to me as if is using a previous version of it.

You're quite right. I had deleted all of the `node_modules` directories and ran `yarn` but I guess this wasn't enough so I recloned the `nightingale` and removed the `react-msa-viewer` repo.

gustavo-salazar:
@dlrice  I think I found the issue. Which comes down to `0==false`. 🤦🏽‍♂️
Check the commit in msa if you are interested: https://github.com/ebi-webcomponents/react-msa-viewer/commit/decf3027892b57d94fce36f998dd8667ce874939

dlrice (approved):

gustavo-salazar:
@xwatkins @aurel-l besides the addition of the `protvista-links` and some fixes in `protvista-msa` this PR also includes a couple of changes I did in the config files of the project. Do you mind having a look to those changes? You can just focus on that part, @dlrice have already code review the new component side of things. 

The config files I changed are:

- [ ] `.eslintrc.json`
- [ ] `package.json`
- [ ] `webpack.config.js`

The one that worries me the most is that I changed the target of the `ts-loader` in the webpack config from `es5` to `es6`, as transpiling it to es5 was causing some weird behaviour in simple stuff such as `for ... of` loops. 
Is having `es6` as target OK with you guys in UniProt?

aurel-l:
> @xwatkins @aurel-l besides the addition of the `protvista-links` and some fixes in `protvista-msa` this PR also includes a couple of changes I did in the config files of the project. Do you mind having a look to those changes? You can just focus on that part, @dlrice have already code review the new component side of things.
> 
> The config files I changed are:
> 
>     * [ ]  `.eslintrc.json`
> 
>     * [ ]  `package.json`
> 
>     * [ ]  `webpack.config.js`
> 
> 
> The one that worries me the most is that I changed the target of the `ts-loader` in the webpack config from `es5` to `es6`, as transpiling it to es5 was causing some weird behaviour in simple stuff such as `for ... of` loops.
> Is having `es6` as target OK with you guys in UniProt?

ES6 target should be fine with us, we should handle that sooner or later. We might need to change our webpack config to make sure to handle that properly for those packages, but make sure to make that a major version bump as we might not be the only ones using those.

I'm just thinking, is that going to be a problem for other people? And for the basic cjs package build?

gustavo-salazar:
Good point about releasing this as a major version.

> I'm just thinking, is that going to be a problem for other people? And for the basic cjs package build?

I'm not sure. I'm also unsure of how webpack deal with the 2 different targets that we have... `js` files are dealt with `babel-loader` with a target for ie11, and `ts` files with `ts-loader` with tharget `e6`.

gustavo-salazar:
So I just created a codepen using canary deployed components from this branch, and everything seems to be working alright:
https://codepen.io/gustavo-salazar/pen/rNxaMgb

aurel-l (approved):

PR #112, opened 2020-10-02
Create the hidelabel option for protvista-msa in the case that con…

dlrice:
…sumers want to provide their own.

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide

### DIFF
--- a/packages/protvista-msa/src/protvista-msa.js
+++ b/packages/protvista-msa/src/protvista-msa.js
@@ -75,10 +75,6 @@ class ProtvistaMSA extends ProtvistaZoomable {
     if (this.hasAttribute("activeLabel")) {
       this.activeLabel = this.getAttribute("activeLabel");
     }
-
-    this._hidelabel = this.hasAttribute("hidelabel")
-      ? this.getAttribute("hidelabel") === "true"
-      : false;
   }
 
   static get observedAttributes() {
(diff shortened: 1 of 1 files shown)


gustavo-salazar:
So I checked the code and seems alright, but if I edit the attribute and leave it empty just like a flag(e.g. `<protvista-msa hidelabel >`), it doesn't work as expected. 

It works if is initialised that way, as somehow it gets rendered as `hidelabel="true"`*:
Screenshot 2020-10-05 at 16 54 56

And if I delete the `="true"` part it displays the labels:
Screenshot 2020-10-05 at 16 57 10

*I'm guessing is related to the `attributeChangedCallback` in the `zoomable` component.

Otherwise it looks OK and even with this issue, it won't affect its use in the interpro side. So I'm happy to approve the request if you think this issue is out of the scope of this PR.

dlrice:
> So I checked the code and seems alright, but if I edit the attribute and leave it empty just like a flag(e.g. `<protvista-msa hidelabel >`), it doesn't work as expected.
> 
> It works if is initialised that way, as somehow it gets rendered as `hidelabel="true"`*:
> Screenshot 2020-10-05 at 16 54 56
> 
> And if I delete the `="true"` part it displays the labels:
> Screenshot 2020-10-05 at 16 57 10
> 
> *I'm guessing is related to the `attributeChangedCallback` in the `zoomable` component.
> 
> Otherwise it looks OK and even with this issue, it won't affect its use in the interpro side. So I'm happy to approve the request if you think this issue is out of the scope of this PR.

Good spot. I have updated the handling of the `hidelabel` attribute to be more consistent by not passing to `observedAttributes`.

gustavo-salazar (approved):

PR #113, opened 2020-11-17
Migrating from travis CI into GitHub actions

gustavo-salazar:
### Reference to issue
Travis has changed their policies. 
1. [LINK](https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom) `.org` will disappear and all OS project will get move to `.com`
2. [LINK](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) change in their paying plans. Giving only a once off set of credits to use in the free plan, once that runs out you need to move into their pay plan.

### Description of changes
I have setup the same test and deploy task that we had in travius but using github actions. You can see the runs coming from the branch here: https://github.com/ebi-webcomponents/nightingale/actions

Also included the readme about creating a new component, which was deploy it in [this run](https://github.com/ebi-webcomponents/nightingale/actions/runs/368263689). And it includes a link to the creating an adapter README, that would hopefully be deployed it when the PR gets merged.

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide


aurel-l (approved):
Nice one 👍

PR #114, opened 2020-11-24
Vcf adapter

xwatkins:
### Reference to issue
Create a new data adapter to load and transform data from VCF files to be viewed in `protvista-variation`.

### Description of changes
- Added _manual_ type definitions to some of the components.
- Created a new component, `protvista-vcf-adapter` to handle VCF files.
  1. Use the `vcftojson` module to convert the files as well as use the Ensembl VEP endpoint to convert genomic to protein positions and residues
  2. Filter out the transcripts from 1. which don't correspond to the current accession number
  3. Use the proteins API to get the sequence
  4. Pass the sequence and converted variant data to `protvista-variation`
- Temporarily modified `ProtvistaVariation.jsx` to demo feature. [Use this file](https://raw.githubusercontent.com/ebi-uniprot/vcf-to-json/master/examples/example_format_1.vcf) as input

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide


packages/protvista-variation-adapter/src/protvista-variation-adapter.ts
```diff
@@ -49,37 +35,27 @@ export const transformData = (
 class ProtvistaVariationAdapter
   extends ProtvistaFeatureAdapter
   implements NightingaleElement {
-  static get is() {
+  static get is(): string {
```
aurel-l, line 38: I believe this type is already defined in Nightingale element. So hopefully we don't need to write it here as it should be implied.
Ideally we'd want all the standard custom element methods to be the same to, no need to type them. But I'm not sure if there's a proper definition with TypeScript. I belive there is `CustomElementConstructor` that would just define the constructor. Maybe we need to have NightingaleElement extend a CustomElement (hopefully already defined somewhere, otherwise we'll have to type it), that would itself extend HTMLElement (which does exist for sure)

packages/protvista-vcf-adapter/src/dataTransformer.ts
```diff
@@ -0,0 +1,71 @@
+/* eslint-disable no-param-reassign */
+import { ProtvistaVariationDatum } from "protvista-variation";
+import { SourceType } from "protvista-variation-adapter/dist/es/variants";
+import { VCFJSON } from "vcftojson/dist/types";
+
+const JSONToHTML = (obj: any, level = 1, ignoreKeys?: string[]): string => {
+  if (typeof obj !== "object") {
+    return `<span>${obj}</span>`;
+  }
+  return `<ul>${Object.entries(obj).reduce((accumulator, [key, value]) => {
```
aurel-l, line 10: What's up with all these `reduce`s lately 😅?

packages/protvista-vcf-adapter/src/dataTransformer.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (ignoreKeys?.includes(key)) {
+      return accumulator;
+    }
+    if (Array.isArray(value)) {
+      const reduced = `${value.reduce(
```
aurel-l, line 15: Wouldn't that be the same than `map` + `.join('')`?

packages/protvista-vcf-adapter/src/protvista-vcf-adapter.ts
```diff
@@ -0,0 +1,90 @@
...
+    this._adaptedData = transformData(vcfJson, this.accession);
+    this._emitEvent();
+  }
+
+  static get is(): string {
+    return "vcf-adapter";
+  }
+
+  onChange(fileSelect: Event): void {
+    const target = (fileSelect.target as unknown) as HTMLInputElement;
```
aurel-l, line 67: Maybe doing an early escape with `if (!(fileSelect.target instanceof HTMLInputElement)) return;` would either set the type of target as HTMLInputElement, or at least let you cast it without resorting to `...as unknown as...`

packages/protvista-variation-adapter/src/protvista-variation-adapter.ts
xwatkins: As an aside, I think we need a place to define abstract classes used by multiple packages (`NightingaleElement` is currently defined in `data-loader`). There's quite a lot of "magic" methods in `protvista-track` and in the data adapters and having a common interface defined for these would be useful. What would be the best here...a package which is just exporting these shared type definitions?

In this specific case, `ProtvistaVariationAdapter` doesn't seem to pick up that the `is` getter corresponds to the readonly attribute in `NightingaleElement`. Actually looking at `DataLoader` it looks like you had to re-define the output type there too...

aurel-l: Yeah, don't really remember what happened when I did it, I was just listing what I wished it would do. I guess we can re-work that when we have a look at Nightingale as a whole next month. I was thinking about an independent type package, or something like that. We'd need to have a look at what's the usual approach for shared resources with Lerna, but we might want to publish just the types themselves as their own package if people want to create their own "Nightingale-compatible" packages outside of our repo.

xwatkins: That sounds like a good plan.

PR #121, opened 2020-12-17
Hackathon 2020 sequence tests

xwatkins:
- Tests are now ran globally instead of using `lerna exec` 🎉
- Added 3 tests to `protvista-sequence` to check:
  - Basic rendering of the component
  - Change of `displaystart` and `displayend`
  - Change of hightlight
- Will move the polyfills defined in `beforeAll` to a common place next

### DIFF
--- a/packages/protvista-zoomable/package.json
+++ b/packages/protvista-zoomable/package.json
@@ -30,7 +30,6 @@
   },
   "gitHead": "3768065ef5b1f65e31acbf52e86cac3d98e347fb",
   "devDependencies": {
-    "@testing-library/dom": "^7.29.0",
     "d3": "^5.7.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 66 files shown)


packages/textarea-sequence/package.json
```diff
@@ -23,5 +23,8 @@
     "lodash-es": "^4.17.11",
     "quill": "1.3.7"
   },
-  "homepage": "https://ebi-webcomponents.github.io/nightingale/"
+  "homepage": "https://ebi-webcomponents.github.io/nightingale/",
+  "devDependencies": {
```
aurel-l, line 27: That probably needs to be hoisted up to the root of the repo if the tests are run globally

packages/protvista-zoomable/package.json
```diff
@@ -30,6 +30,7 @@
   },
   "gitHead": "3768065ef5b1f65e31acbf52e86cac3d98e347fb",
   "devDependencies": {
+    "@testing-library/dom": "^7.29.0",
```
aurel-l, line 33: Same thing

aurel-l (approved):

Issue #128, opened 2021-01-21
Zoom buttons


xwatkins:
**Describe the usability issue**
Usability tests have shown that it is not obvious to users that it is possible to zoom in and out using `protvista-navigation`.

**Describe the solution that has been discussed by the team**
- Add zoom in and zoom out buttons. These could use a magnifying glass icon and (-) (+)
- Zoom could happen by a certain number of steps based on the sequence length
- There could additionally be a "Zoom to sequence" button (TBD) to zoom all the way to the sequence level

**How does the solution address the problem?**
Clear buttons and iconography associated with the concept of "zoom" will help advertise the functionality to users

**Additional context**
Should this be part of `protvista-navigation`, its own component, or implemented in the client? Keeping in mind that it might be displayed in different places depending on who is using the components.

PR #133, opened 2021-02-05
Use slots and css variables

xwatkins:
### Reference to issue
#128 

### Description of changes
- For the content of `<button/>`, change to using slots
- css to reset the button styling and apply default styles, and expose some properties through css variables
- Add an option (through a slot) to provide an extra button to zoom in right to the sequence level (30 amino acids)
- Refactor the zooming logic to only move the `end` unless it's reached the full length

These changes will have implications with InterPro's use of the component @Swaathik @gustavo-salazar 

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide

### DIFF
--- a/packages/protvista-zoom-tool/src/ProtvistaZoomTool.js
+++ b/packages/protvista-zoom-tool/src/ProtvistaZoomTool.js
@@ -17,7 +17,7 @@ class ProtvistaZoomTool extends HTMLElement {
       parseFloat(this.getAttribute("displayend")) || this._length;
     this._scaleFactor =
       parseFloat(this.getAttribute("scalefactor")) || this._length / 5;
-
+    this._zoomToSeq = this.querySelector('[slot="zoom-in-seq"]') !== null;
     this.renderContent();
   }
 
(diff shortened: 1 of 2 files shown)


packages/protvista-zoom-tool/src/ProtvistaZoomTool.js
```diff
@@ -26,75 +32,83 @@ class ProtvistaZoomTool extends HTMLElement {
...
+    let k = 0;
+    if (operation === "zoom-in") {
+      k = this._scaleFactor;
+    } else if (operation === "zoom-out") {
+      k = -this._scaleFactor;
+    } else if (operation === "zoom-in-seq") {
+      k =
+        this._displayend -
+        this._displaystart -
+        (this._displaystart === 1 ? 29 : 30);
```
gustavo-salazar, line 44: Why is this necessary?

gustavo-salazar (approved):
Happy with this changes. Just wondering why to make the difference between 29 or 30 residues in the line I marked?

PR #135, opened 2021-03-10
Nightingale linegraph track

Swaathik:
### Reference to issue
Replacement for Protvista-variation-graph for it to be more generic

### Description of changes
- It accepts multiple charts
- Colour customisation. If not specified, d3 colour scale is used.
- Range of min-max values. 
- Allow the use of the (or some of the) different line curve types defined in the D3 API (https://github.com/d3/d3-shape/blob/v2.0.0/README.md#curves)
- Allow selection of a data point. 

### Tests / Styleguide
 - [ ] Tests pass
 - [ ] Styleguide

### DIFF
--- a/packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
+++ b/packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
@@ -99,16 +99,11 @@ class NightingaleLinegraphTrack extends ProtvistaTrack {
         d3.selectAll(".mouse-per-line circle").style("opacity", "0");
         d3.selectAll(".mouse-per-line text").style("opacity", "0");
       })
-      // .on('mouseover', function() { // on mouse in show circles and text
-      //   d3.selectAll(".mouse-per-line circle")
-      //       .style("opacity", "1");
-      //   d3.selectAll(".mouse-per-line text")
-      //       .style("opacity", "1");
-      // })
       .on("mousemove", () => {
         // mouse moving over canvas
         const mouse = d3.mouse(this);
 
+        // Showing the circle and text only when the mouse is moving over the paths
         if (
           mouse[0] < _this.xScale(_this.beginning) ||
           mouse[0] > _this.xScale(_this.end) + _this.getSingleBaseWidth()
@@ -119,12 +114,12 @@ class NightingaleLinegraphTrack extends ProtvistaTrack {
           d3.selectAll(".mouse-per-line circle").style("opacity", "1");
           d3.selectAll(".mouse-per-line text").style("opacity", "1");
 
-          const features = [];
+          const features = {};
           const seqPosition = Math.floor(_this.xScale.invert(mouse[0]));
 
           d3.selectAll(".mouse-per-line text").text((d) => {
             const value = d.values.find((v) => v.position === seqPosition);
-            features.push(value);
+            features[d.name] = value;
             return value ? value.value : "";
           });
 
@@ -133,6 +128,10 @@ class NightingaleLinegraphTrack extends ProtvistaTrack {
             let end = lines.nodes()[i].getTotalLength();
             let target = null;
             let pos = {};
+            /*
+             Finding the nearest point in the path to the mouse pointer using iterative dichotomy.
+             Example can be found here - https://bl.ocks.org/larsenmtl/e3b8b7c2ca4787f77d78f58d41c3da91
+             */
             while (true) {
               target = Math.floor((beginning + end) / 2);
               pos = lines.nodes()[i].getPointAtLength(target);
--- a/packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
+++ b/packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
@@ -37,10 +37,25 @@ class NightingaleLinegraphTrack extends ProtvistaTrack {
 
   refresh() {
     if (!this.svg) return;
-    this.svg.selectAll("path").remove();
+    this.svg.selectAll("g.chart-group").remove();
+    const chartGroup = this.svg.append("g").attr("class", "chart-group");
     this._initScales();
 
-    this._data.map((d) => this.drawLine(d));
+    chartGroup
+      .selectAll(".chart")
+      .data(this._data)
+      .enter()
+      .append("path")
+      .attr("class", "chart")
+      .attr("id", (d) => d.name)
+      .attr("d", (d) => {
+        this.drawLine(d);
+        return this.line(d.values);
+      })
+      .attr("fill", "none")
+      .attr("stroke", (d) => d.colour || interpolateRainbow(Math.random()))
+      .attr("transform", "translate(0,0)")
+      .call(this.bindEvents, this);
     this._updateHighlight();
   }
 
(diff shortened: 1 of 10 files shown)


packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  set data(data) {
+    this._data = data;
+    this._createTrack();
+  }
+
+  _createTrack() {
+    select(this).selectAll("svg").remove();
+    this.svg = select(this)
```
xwatkins, line 30: I know this is in the old component, but a lot of this is already present in protvista track, so we could just use `super._createTrack()`. However be careful to call it after the bit of code where max/min ranges are defined

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -0,0 +1,149 @@
...
+    const range = [];
+    this._data.map((d) => range.push(...d.range));
+    this.minRange = min(range);
+    this.maxRange = max(range);
+
+    // Create the visualisation here
+    this.refresh();
+  }
+
+  refresh() {
```
xwatkins, line 45: Looking at the old component, I think the use of `refresh` was wrong, as the idea is to use `_createTrack` to add elements (d3's enter) and `refesh` is only supposed to update (anything after the `enter()`). This is so that the elements are only added once, and the only thing that happens is that they are updated when `refresh` is called: currently they are removed and re-added, which causes performance issues. Let me know if it's unclear and we can look at it together.

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -0,0 +1,149 @@
...
+  _initYScale() {
+    this._yScale
+      .domain([this.minRange, this.maxRange])
+      .range([this._height, 0]);
+  }
+
+  drawLine(d) {
+    this._curve = d.lineCurve || "curveLinear";
+
+    this.line = line()
```
xwatkins, line 96: I think it's dangerous to set this line as an attribut of the object as it's updated in every loop. You could get into some issues with context. This `drawLine` function could actually just be a function outside of the object which returns a line object.

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -0,0 +1,149 @@
...
+    this.dispatchEvent(
+      new CustomEvent("change", {
+        detail,
+        bubbles: true,
+        cancelable: true,
+      })
+    );
+  }
+
+  _checkSeqBaseWidth() {
```
xwatkins, line 127: I'm a bit confused by this function, what does it do?

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -0,0 +1,149 @@
...
+    const xratio = 0.8;
+    const tempNode = this.svg.append("text").attr("class", "base").text("T");
+    const chWidth = tempNode.node().getBBox().width * xratio;
+    tempNode.remove();
+    const ftWidth = this.getSingleBaseWidth();
+    this.isSeqBaseVisible = ftWidth - chWidth > 1;
+  }
+
+  addPointer(coords) {
+    this.chart.selectAll("circle.pointer").remove();
```
xwatkins, line 137: Actually wondering if it would be better to have small semi-transparent circles always visible at the end of lines, and have the event be attached to these rather than the lines?

xwatkins (changes requested):

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
Swaathik: super._createTrack() calls this https://github.com/ebi-webcomponents/nightingale/blob/master/packages/protvista-track/src/protvista-track.js#L150. Since there are no features it will fail. Please correct me if I am wrong

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
Swaathik: I understand what you mean. I have updated the code in the createTrack to do all the initial creation and assigning data. Refresh only redraws the line with the latest x and y values

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
Swaathik: `this` was stupid of me. I have changed the `drawLine` to just return a line object.

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
Swaathik: This decides whether the residues are shown above (similar to protvista-sequence), so that the highlight on mouseover falls exactly at the right position. If not, when it is zoomed out, the highlight always is visible at the end of the graph

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
Swaathik: Could you explain it? Meaning having circles at the data points and attaching the events to the circles?

xwatkins: I was saying that because I noticed a performance issue when mousing over the lines. I think this is happening because the event keeps getting triggered over and over again and it causes a backlog?
Actually a better approach might be something like [this](https://bl.ocks.org/larsenmtl/e3b8b7c2ca4787f77d78f58d41c3da91)?

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
xwatkins: You're correct - I'm happy with the logic being taking out of `_refresh` so all good.

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
Swaathik: Thank you. It looks very cool. I will try to have something like this.

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -0,0 +1,197 @@
...
+            features.push(value);
+            return value ? value.value : "";
+          });
+
+          d3.selectAll(".mouse-per-line").attr("transform", (d, i) => {
+            let beginning = 0;
+            let end = lines.nodes()[i].getTotalLength();
+            let target = null;
+            let pos = {};
+            while (true) {
```
xwatkins, line 136: Could you add some comments about what this is doing (trying to find the closest point to the mouse pointer in the paths by using iterative dichotomy) and the link to the d3 example?

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -0,0 +1,197 @@
...
+      .attr("width", this.width) // can't catch mouse events on a g element
+      .attr("height", this._height)
+      .attr("fill", "none")
+      .attr("pointer-events", "all")
+      .on("mouseout", () => {
+        // on mouse out hide circles and text
+        d3.selectAll(".mouse-per-line circle").style("opacity", "0");
+        d3.selectAll(".mouse-per-line text").style("opacity", "0");
+      })
+      // .on('mouseover', function() { // on mouse in show circles and text
```
xwatkins, line 102: Did you want to leave this commented out on purpose for later, or can it be removed?

xwatkins (approved):
Just a comment about adding more comments 😄

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
Swaathik: Hi Xavier, yes, it is doing the same. I will add the comment and the link too.

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
Swaathik: It has to be removed. Mousemove does the job. It is redundant

PR #138, opened 2021-04-14
Add support for Ensembl COVID-19 links

xwatkins:
### Reference to issue
Some variants now have extra information allowing to link to Ensembl COVID-19. This should be displayed in the tooltip.

### Description of changes
Add logic to extract the corresponding id and generate the link.

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide

### DIFF
--- a/packages/protvista-variation-adapter/src/tooltipGenerators.ts
+++ b/packages/protvista-variation-adapter/src/tooltipGenerators.ts
@@ -46,6 +46,20 @@ export const getPopulationFrequencies = (
     )
     .join("")}`;
 
+export const getEnsemblCovidLinks = (variant: Variant) => {
+  const shouldGenerateLink = variant.locations.some(
+    (location) => location.source === "EnsemblViruses"
+  );
+  if (shouldGenerateLink) {
+    const { id } = variant.xrefs.find((xref) => xref.name === "ENA");
+    return (
+      `<h5>Ensembl COVID-19</h5>` +
+      `<p><a href="https://covid-19.ensembl.org/Sars_cov_2/Variation/Explore?v=${id}" target="_blank" rel="noopener noreferrer">${id}</a></p>`
+    );
+  }
+  return "";
+};
+
 export const getPredictions = (predictions: Prediction[]) => {
   const groupedPredictions = groupBy(predictions, "predAlgorithmNameType");
   const counts = Object.keys(groupedPredictions).map((key) => {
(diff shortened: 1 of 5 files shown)


packages/protvista-variation-adapter/src/tooltipGenerators.ts
```diff
@@ -46,6 +46,20 @@ export const getPopulationFrequencies = (
     )
     .join("")}`;
 
+export const getEnsemblCovidLinks = (variant: Variant) => {
+  const shouldGenerateLink = variant.locations.some(
+    (location) => location.source === "EnsemblViruses"
```
aurel-l, line 51: Is this always going to be in Ensembl covid 19 portal? "EnsemblViruses" feels like it's going to cover more than just covid 19

xwatkins: Yes, confirmed by Jie. Also note the "interesting" way to get the id. 😨

aurel-l: Sure... Because why not? 😅  (Similar to what we had at some point in the citations I think, where the pubmeds and doi were xrefs in a list...).

packages/protvista-variation-adapter/src/tooltipGenerators.ts
```diff
@@ -46,6 +46,20 @@ export const getPopulationFrequencies = (
...
+  );
+  if (shouldGenerateLink) {
+    const { id } = variant.xrefs.find((xref) => xref.name === "ENA");
+    return (
+      `<h5>Ensembl COVID-19</h5>` +
+      `<p><a href="https://covid-19.ensembl.org/Sars_cov_2/Variation/Explore?v=${id}" target="_blank" rel="noopener noreferrer">${id}</a></p>`
+    );
```
aurel-l, line 58: I would do
```suggestion
    const xref = variant.xrefs.find((xref) => xref.name === "ENA");
    return xref.id ? (`
      <h5>Ensembl COVID-19</h5>
       <p><a href="https://covid-19.ensembl.org/Sars_cov_2/Variation/Explore?v=${id}" target="_blank" rel="noopener noreferrer">${id}</a></p>
    `) : '';
```
Just in case the xref doesn't exist, or doesn't have an ID

xwatkins: Done (but not through suggestion)

aurel-l (approved):

PR #139, opened 2021-04-15
Contact map

Swaathik:
### Reference to issue
A heatmap visualisation for Alignment's contact data. 

### Description of changes
A new package called Nightingale contact map is created.

### Tests / Styleguide
 - [ ] Tests pass
 - [ ] Styleguide


gustavo-salazar:
I like the way it looks. I'll check the code for `packages/nightingale-contact-map/src/nightingale-contact-map.js` in more detail in a minute... but I just wanted to add 2 general comments:
1. We are trying to get the names and components themselves more generic. so instead of `contacts-map` something as `heatmap` might be better.
2. I'm worry about the performance. the example creates >20k rect DOM nodes, wouldn't this be better using canvas instead of SVG?

packages/nightingale-heatmap/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+The height specifies the height of the heat map.
+
+#### `data: Array`
+
+The data array is of the following structure.
+
+```javascript
+[
+  [1, 1, 0.2],
```
gustavo-salazar, line 33: Is the data sparse? So what color would appear in position 2,1?

gustavo-salazar (approved):
I thought you were making it that the rectangle that the mouse is currently hovering would be highlighted, but thats not the case in: https://deploy-preview-139--nightingale-elements.netlify.app/#/heatmap
I can see the events are triggered properly so I guess the highlight is just a nice to have?

In any case I'm happy with the PR.

packages/nightingale-heatmap/README.md
Swaathik: In this case, it is symmetric. I have values only for one half. Speaking about that, I can add an attribute asking if it is symmetric and add the symmetric half depending on that.

gustavo-salazar (approved):

packages/nightingale-heatmap/README.md
gustavo-salazar: Cool. I was also going for the possibility of default values. For instance  in a 5x5 matrix the data could be only 1 point`[ [2, 4, 0.8] ]` and then everything else is a `0`. NO need to implement this, just thought of it as a possibility once I saw this structure

PR #141, opened 2021-05-02
Extended highlighting capability (color)

scheuerv:
### Description of changes
I'm working on a new version of [MolArt](https://github.com/scheuerv/typed-molart) and I need to have highlights of different colors for better visualization, so I propose the following improvement.

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide

### DIFF
--- a/packages/protvista-utils/src/Region.js
+++ b/packages/protvista-utils/src/Region.js
@@ -23,14 +23,15 @@ export default class Region {
       this.segments = [];
       return;
     }
-    this.segments = this.regionString.split(",").map(region => {
-      const [_start, _end, _] = region.split(":");
+    this.segments = this.regionString.split(",").map((region) => {
+      const [_start, _end, _color, _] = region.split(":");
       if (typeof _ !== "undefined")
         throw new Error(
-          `there should be at most 1 ':' per region. Region: ${region}`
+          `there should be at most 2 ':' per region. Region: ${region}`
         );
       let start = _start ? Number(_start) : this.min;
       let end = _end ? Number(_end) : this.max;
+      const color = _color;
       if (start > end) [start, end] = [end, start];
       if (start < this.min) start = this.min;
       if (end > this.max) end = this.max;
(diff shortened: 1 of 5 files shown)


packages/protvista-utils/src/Region.js
```diff
@@ -8,11 +8,11 @@ export default class Region {
 
   encode(full = false) {
     return this.segments
-      .map(({ start, end }) => {
-        if (full) return `${start}:${end}`;
+      .map(({ start, end, color }) => {
+        if (full) return `${start}:${end}:${color ?? "Default"}`;
```
xwatkins, line 12: Could color always be optional?
```suggestion
        if (full) return `${start}:${end}${color ? `:${color}` : ""}`;
```

packages/protvista-utils/src/TrackHighlighter.js
```diff
@@ -8,6 +8,7 @@ const combineRegions = (region1, region2) => {
 export default class TrackHighlighter {
   constructor({ element, min, max }) {
     this.element = element;
+    this.element._highlightcolor = "#FFEB3BCC";
```
xwatkins, line 11: This could be set as a `const` at the top of the file:

`const DEFAULT_HIGLIGHT_COLOR="#FFEB3BCC"`

packages/protvista-utils/src/TrackHighlighter.js
```diff
@@ -25,12 +26,18 @@ export default class TrackHighlighter {
...
+      this.element._highlightcolor = this.element.getAttribute(
+        "highlightcolor"
+      );
       if (
         this.element._highlightstart !== null &&
         this.element._highlightend !== null &&
         typeof this.element._highlightstart === "number" &&
         typeof this.element._highlightend === "number"
       ) {
+        this.element._highlightcolor = this.element._highlightcolor
```
xwatkins, line 38: You might be able to remove this line and use the shorthand L29 above:

```
this.element._highlightcolor = this.element.getAttribute(
        "highlightcolor"
      ) || DEFAULT_HIGLIGHT_COLOR
```

packages/protvista-utils/src/TrackHighlighter.js
```diff
@@ -89,14 +96,14 @@ export default class TrackHighlighter {
       .enter()
       .append("rect")
       .style("opacity", 0.5)
```
xwatkins, line 98: I wonder if we still need this line with the alpha defined in the hex color. Might need to drop the value a little in the hex default value definition to match the current color (or slightly darker, as it's sometimes difficult to see for variants)

xwatkins (changes requested):
Thanks for the PR! A couple of comments, but looks good 👍

xwatkins (approved):

PR #148, opened 2021-05-11
Filled area chart using "area" field in line plot

prihoda:
Use `fill: true` in line plot to produce a filled area chart

Implements https://github.com/ebi-webcomponents/nightingale/issues/149

image

### DIFF
--- a/packages/nightingale-linegraph-track/README.md
+++ b/packages/nightingale-linegraph-track/README.md
@@ -32,8 +32,7 @@ The data expects the following structure.
     range:[min, max],
     colour?: Line colour,
         (Colour will be assigned if not provided. Use "none" for no line colour)
-    fill?: Produce a filled area chart (default false),
-    fillColor?: Fill colour (same as line colour if not provided),
+    fill?: Create area plot using given fill colour (default "none"),
     lineCurve?: 'curveLinear'(default)|'curveBasis'|'curveCardinal'|'curveStep'|'curveNatural',
         (More curves - https://github.com/d3/d3-shape/blob/v2.0.0/README.md#curves)
     values: [
--- a/packages/nightingale-linegraph-track/README.md
+++ b/packages/nightingale-linegraph-track/README.md
@@ -31,6 +31,7 @@ The data expects the following structure.
     name: String,
     range:[min, max],
     colour?: Any colour,
+    area?: Produce a filled area chart (default false),
     lineCurve?: 'curveLinear'(default)|'curveBasis'|'curveCardinal'|'curveStep'|'curveNatural',
         (More curves - https://github.com/d3/d3-shape/blob/v2.0.0/README.md#curves)
     values: [
--- a/packages/nightingale-linegraph-track/README.md
+++ b/packages/nightingale-linegraph-track/README.md
@@ -30,8 +30,10 @@ The data expects the following structure.
 {
     name: String,
     range:[min, max],
-    colour?: Any colour,
-    area?: Produce a filled area chart (default false),
+    colour?: Line colour,
+        (Colour will be assigned if not provided. Use "none" for no line colour)
+    fill?: Produce a filled area chart (default false),
+    fillColor?: Fill colour (same as line colour if not provided),
     lineCurve?: 'curveLinear'(default)|'curveBasis'|'curveCardinal'|'curveStep'|'curveNatural',
         (More curves - https://github.com/d3/d3-shape/blob/v2.0.0/README.md#curves)
     values: [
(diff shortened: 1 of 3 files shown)


packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -192,14 +192,23 @@ class NightingaleLinegraphTrack extends ProtvistaTrack {
   drawLine(d) {
     const curve = d.lineCurve || "curveLinear";
 
-    return line()
+    var graph;
```
xwatkins, line 195: Please use `let`|`const`
```suggestion
    let graph;
```

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -192,14 +192,23 @@ class NightingaleLinegraphTrack extends ProtvistaTrack {
...
+    if (d.area) {
+      graph = area()
+        .y1((d) => this._yScale(d.value))
+        .y0((d) => this._yScale(0))
```
xwatkins, line 199:
```suggestion
        .y0(() => this._yScale(0))
```

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -59,8 +59,8 @@ class NightingaleLinegraphTrack extends ProtvistaTrack {
         d.colour = d.colour || interpolateRainbow(Math.random()); // eslint-disable-line no-param-reassign
         return this.drawLine(d)(d.values);
       })
-      .attr("fill", "none")
-      .attr("stroke", (d) => d.colour)
+      .attr("fill", (d) => d.area ? d.colour : "none")
```
xwatkins, line 62: Could there be instances where we want a different stroke and fill colour? In that case maybe we should add a `fillColor` attribute to the configuration object, and this could be the flag to use instead of `area`?

xwatkins (changes requested):

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
prihoda: Good thinking. And is it `fillColor` or `fillColour` ? ;) I would personally change it to `color` and `fillColor`, and keep the `colour` back-compatible.

xwatkins: 👍  It should be `fillColor` - we're trying to stick to US spelling although a codebase-wide search/replace might be needed at some point 😄

prihoda: ok so it will be `fillColor` and `colour` for now?

xwatkins: Sounds good, we'll update later

packages/nightingale-linegraph-track/src/nightingale-linegraph-track.js
```diff
@@ -57,9 +57,10 @@ class NightingaleLinegraphTrack extends ProtvistaTrack {
       .attr("id", (d) => d.name)
       .attr("d", (d) => {
         d.colour = d.colour || interpolateRainbow(Math.random()); // eslint-disable-line no-param-reassign
+        d.fillColor = d.fillColor || d.colour
         return this.drawLine(d)(d.values);
       })
-      .attr("fill", "none")
+      .attr("fill", (d) => d.fill ? d.fillColor : "none")
```
xwatkins, line 63:
```suggestion
      .attr("fill", (d) => d.fillColor ? d.fillColor : "none")
```

packages/nightingale-linegraph-track/README.md
```diff
@@ -30,7 +30,10 @@ The data expects the following structure.
 {
     name: String,
     range:[min, max],
-    colour?: Any colour,
+    colour?: Line colour,
+        (Colour will be assigned if not provided. Use "none" for no line colour)
+    fill?: Produce a filled area chart (default false),
```
xwatkins, line 35: For the sake of simplicity I was wondering if we could remove the `fill` attribute and use `fillColor` as the flag instead (see suggestion below)? This would mean not having a default `fillColor`.

xwatkins (changes requested):
Looks good, just one small change. Thank you!

packages/nightingale-linegraph-track/README.md
prihoda: yeah that also crossed my mind... so since you say that absence of default fillColor is not an issue, I'll make that change :)

prihoda: or what about just using `fill` as the single attribute? `fill: "red"` looks intuitive

xwatkins: That way we get rid of the "color" vs "colour" issue!

xwatkins (approved):

prihoda:
Thanks @xwatkins, anything required from me at this stage?

PR #154, opened 2021-06-07
Molstar

dlrice:
### Reference to issue
Move from using Litemol in protvista-structure to using Mol* ([Jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-26052))

### Description of changes

- Create wrapper `structure-viewer.ts` to act as intermediary between Nightingale and Mol*
- Converted JavaScript files to TypeScript
- Mol* added as dependency rather than peer dependency as the consumer of protvista-structure is uniprot.org which will not have react.
- React added as a dependency as Mol* raises issues without this

### Tests / Styleguide
 - [ ] Tests pass
 - [ ] Styleguide

### Note
- Seeing some issues in translating the coordinates - looking into this
- `yarn build:es` must precede `yarn start`
- If you're having issues getting the demo app to work do:
  ```
  rm -rf node_modules
  yarn
  yarn build:es
  yarn start
  ```


packages/protvista-structure/src/position-mapping.ts
```diff
@@ -0,0 +1,87 @@
...
+  mappings: Mappings,
+  mappingDirection: Direction = "UP_PDB"
+): {
+  start: number;
+  end: number;
+  entity: number;
+  chain: string;
+} | null => {
+  // return if they have been set to 'undefined'
+  if (!start || !end || Number.isNaN(start) || Number.isNaN(end)) {
```
dlrice, line 46: There is an additional check in the [original file](https://github.com/ebi-webcomponents/nightingale/blob/master/packages/protvista-structure/src/protvista-structure.js#L304):
```
if (
  typeof this.highlight === "string" ||  <--- do we need this in translatePositions or check it before calling translatePositions
  Number.isNaN(start) ||
  Number.isNaN(end)
) {
  return null;
}
```

packages/protvista-structure/package.json
```diff
@@ -25,11 +29,14 @@
...
   "dependencies": {
-    "lodash-es": "^4.17.11",
-    "whatwg-fetch": "^3.4.1"
-  },
-  "peerDependencies": {
-    "litemol": "^2.4.2"
+    "data-loader": "^3.0.0",
+    "molstar": "2.0.6",
+    "react": "17.0.2",
+    "whatwg-fetch": "3.4.1"
```
xwatkins, line 35: As we have a dependency on `data-loader`, might as well use it for the fetch too and not depend on `whatwg-fetch`?

packages/protvista-structure/src/protvista-structure.ts
```diff
@@ -0,0 +1,309 @@
+/* eslint-disable class-methods-use-this */
+import "whatwg-fetch";
```
xwatkins, line 2: See comment here https://github.com/ebi-webcomponents/nightingale/pull/154/files#diff-bade8e4344307a64705e75ba253da08c5c7cc0d1e8076f0bf8770ba8e48cfe8aR35

packages/protvista-structure/src/protvista-structure.ts
```diff
@@ -0,0 +1,309 @@
...
+  attributeChangedCallback(
+    attrName: string,
+    oldVal: string,
+    newVal: string
+  ): void {
+    if (oldVal !== newVal) {
+      const value = parseFloat(newVal);
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore 🤷🏼‍♂️
+      this[`_${attrName}`] = typeof value === "number" ? newVal : value;
```
xwatkins, line 169: Not sure what this is for as the possible attributes are updated in the code below, and none of them are a number (I think height can be a string because of the units, but if not the `number` check could happen in that block)

packages/protvista-structure/src/protvista-structure.ts
```diff
@@ -0,0 +1,309 @@
...
+      bubbles: true,
+      cancelable: true,
+    });
+    this.dispatchEvent(event);
+  }
+
+  highlightChain(): void {
+    this._structureViewer.clearMessages();
+    if (!this._highlight) {
+      this._structureViewer.clearMessages();
```
xwatkins, line 282: Don't think you need to clear the messages twice?

packages/protvista-structure/src/protvista-structure.ts
```diff
@@ -0,0 +1,309 @@
...
+      return;
+    }
+
+    let translatedPositions;
+    try {
+      translatedPositions = this._highlight
+        .map(({ start, end }) =>
+          translatePositions(start, end, this._selectedMolecule.mappings)
```
xwatkins, line 290: Just to make it easier to read, wondering if it would be worth specifying the direction even though it's the default

packages/protvista-structure/src/position-mapping.ts
xwatkins: Checking that it's the correct type could be done here https://github.com/ebi-webcomponents/nightingale/pull/154/files#diff-ec9fd2d846b88c27b3683134bfbc0eea81436cf17c477b8b68dc97fd8ef273fcR146 or in the `attributeChangedCallback`

xwatkins (changes requested):

packages/protvista-structure/src/protvista-structure.ts
dlrice: I can see two attributes from the README that aren't accounted for in the switch statement:

- [hide-table](https://github.com/ebi-webcomponents/nightingale/blob/master/packages/protvista-structure/README.md#hide-table)
- [hide-viewport-controls](https://github.com/ebi-webcomponents/nightingale/blob/master/packages/protvista-structure/README.md#hide-viewport-controls)

I don't think they are implemented anymore - do we want them to be implemented?

xwatkins: I don't think so - without them being in the list of observed attributes they won't trigger the `attributChangedCallback` anyway

xwatkins: I think these are leftover properties from when Mahdi was trying to add it to PepVep but he ended up doing something else and must have left them in the README by accident

packages/protvista-structure/src/protvista-structure.ts
dlrice: Updated

packages/protvista-structure/src/protvista-structure.ts
dlrice: Debugging relic - removed

packages/protvista-structure/src/position-mapping.ts
dlrice: Added a check in the `highlightChain` before `translatePositions` is called.

packages/protvista-structure/src/protvista-structure.ts
dlrice: 👍 removed

packages/protvista-structure/package.json
dlrice: Done

xwatkins (approved):

PR #157, opened 2021-07-05
Datatable ts

xwatkins:
### Reference to issue
Convert `protvista-datatable` to TypeScript 

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide


packages/protvista-datatable/src/protvista-datatable.ts
```diff
@@ -1,26 +1,79 @@
...
+
+import { ProtvistaLoadEvent } from "./types/events";
+import { ProtvistaManager } from "./types/manager";
+
 import styles from "./styles";
 
-class ProtvistaDatatable extends LitElement {
+type DataTableDatum = {
+  start?: number;
+  begin?: number;
```
dlrice, line 12: I think here you need at least one of these? If so, type-fest has a [require-at-least-one](https://github.com/sindresorhus/type-fest/blob/main/source/require-at-least-one.d.ts) that could enforce that.

packages/protvista-datatable/src/protvista-datatable.ts
```diff
@@ -48,32 +101,40 @@ class ProtvistaDatatable extends LitElement {
...
-    ) {
+  eventHandler(e: MouseEvent): void {
+    const target = e.target as HTMLElement;
+    if (!target.closest("protvista-datatable") && !target.closest(".feature")) {
       this.selectedid = null;
     }
   }
 
-  static get properties() {
+  static get properties(): { [key: string]: any } {
```
dlrice, line 121: Is it possible to eliminate the `any` here?

packages/protvista-datatable/src/protvista-datatable.ts
```diff
@@ -1,26 +1,91 @@
...
+  LitElement,
+  html,
+  TemplateResult,
+  CSSResult,
+  PropertyDeclarations,
+} from "lit-element";
 import { v1 } from "uuid";
 import { ScrollFilter } from "protvista-utils";
-/* eslint-disable import/extensions, import/no-extraneous-dependencies */
+// Not sure why, but eslint doesn't see it...
```
dlrice, line 10: With the following in `.eslintrc.json`:
```
  "settings": {
    "import/resolver": {
      "node": {
        "extensions": [".js", ".ts",".d.ts"]  // <---with ".d.ts" now
      }
    }
  }
```
the linter shouldn't complain anymore.

xwatkins: You, sir, are a gentleman and a scholar. I'd assumed ".ts" would have covered ".d.ts", but no.

dlrice (approved):

PR #159, opened 2021-07-20
Use slot instead of data

xwatkins:
### Reference to issue
The `protvista-datatable` component doesn't allow flexibility in the rendering and actions of its data content.

### Description of changes
Change of strategy in the way `protvista-datable` works to use `slots` instead. This leaves the client more flexibility in rendering the table content.

See https://github.com/ebi-webcomponents/nightingale/blob/6780ea206014029fa209d619e497f91c335d47c0/packages/protvista-datatable/README.md for new definition

### Tests / Styleguide
 - [x] Tests pass
 - [x] Styleguide


packages/protvista-datatable/src/protvista-datatable.ts
```diff
@@ -224,39 +206,58 @@ class ProtvistaDatatable<T extends DataTableDatum> extends LitElement {
...
-    ) {
-      className = `${className} overlapped`;
-    }
-    return className;
-  }
+  updateRowStyling(): void {
+    let oddOrEvenCount = 0;
+    this.rows?.forEach((row) => {
+      const { start, end } = row.dataset;
+      row.classList.add(oddOrEvenCount % 2 === 0 ? "odd" : "even");
```
aurel-l, line 213: Is this because you add rows in the middle for the extra content that you cannot do that only with CSS? Not sure if you'll filter rows at some point, but I guess you'll have to remove any existing class and re-set it whenever you do, o whenever the source table is modified

aurel-l: Also, the end result will be the same just inverted, but the code should probably be `oddOrEvenCount % 2 === 0 ? "even" : "odd"` in order to match what the logic says

xwatkins: > Is this because you add rows in the middle for the extra content that you cannot do that only with CSS?  
Yes. Can't combine `nth-child` with `not` because of CSS's fully declarative nature.
And you're correct about having to update it after filtering.

> in order to match what the logic says
Will update the logic to be the right way round

app/src/components/ProtvistaDatatable.jsx
```diff
@@ -1,86 +1,77 @@
...
-            length="770"
-            layout="non-overlapping"
-          ></protvista-track>
-          <protvista-datatable id="data-table"></protvista-datatable>
-        </protvista-manager>
-      </Fragment>
-    );
-  }
-}
+  // console.log(data);
```
dlrice, line 32:
```suggestion
```

PR #168, opened 2021-10-07
Taxonomy sunburst

gustavo-salazar:
### Description of changes
New component `nightingale-sunburst` to display taxonomies

### DIFF
--- a/packages/nightingale-sunburst/src/nightingale-sunburst.js
+++ b/packages/nightingale-sunburst/src/nightingale-sunburst.js
@@ -8,17 +8,16 @@ import {
   interpolateRainbow,
 } from "d3";
 
+const superkingdoms = ["bacteria", "viruses", "archaea", "eukaryota", null];
+
 const partition = (data, radius, attribute) => {
   const h = hierarchy(data).sort((a, b) => b[attribute] - a[attribute]);
   return d3partition().size([2 * Math.PI, radius])(h);
 };
 
-const getColor = (d, colorFn, nameAttribute) => {
-  if (d.depth === 1) {
-    return colorFn(d.data[nameAttribute]);
-  }
-  return getColor(d.parent, colorFn, nameAttribute);
-};
+const colorFn = scaleOrdinal(
+  quantize(interpolateRainbow, superkingdoms.length + 1)
+).domain(superkingdoms.map((_, i) => i));
 
 const getValue = (d, attributeName) => {
   if (d[attributeName]) return d[attributeName];
@@ -60,7 +59,7 @@ const getLineageFromNode = (node, attributeName, attributeID) => {
     ];
   }
   return [
-    ...getLineageFromNode(node.parent, attributeName),
+    ...getLineageFromNode(node.parent, attributeName, attributeID),
     {
       name: node.data[attributeName],
       id: node.data[attributeID],
@@ -79,7 +78,7 @@ class NightingaleSunburst extends LitElement {
     "weight-attribute": { type: String },
     "weight-attribute-label": { type: String },
     "name-attribute": { type: String },
-    "id-attribute": { type: Number },
+    "id-attribute": { type: String },
     "max-depth": { type: Number },
     "show-label": { type: Boolean },
   };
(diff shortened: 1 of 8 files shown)


packages/nightingale-sunburst/src/nightingale-sunburst.js
```diff
@@ -0,0 +1,358 @@
...
+} from "d3";
+
+const superkingdoms = ["bacteria", "viruses", "archaea", "eukaryota", null];
+
+const partition = (data, radius, attribute) => {
+  const h = hierarchy(data).sort((a, b) => b[attribute] - a[attribute]);
+  return d3partition().size([2 * Math.PI, radius])(h);
+};
+
+const getValue = (d, attributeName) => {
```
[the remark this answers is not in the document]
gustavo-salazar: yeah. It calculates the value by adding the value of its children and it does it recursively. Some of the trees I'm dealing with had value only on the species; other had it all along except for the root node.

packages/nightingale-sunburst/src/nightingale-sunburst.js
```diff
@@ -0,0 +1,358 @@
...
+  if (d[attributeName]) return d[attributeName];
+  if (!d.children) return 0;
+  return d.children.reduce(
+    (agg, item) => agg + getValue(item, attributeName),
+    0
+  );
+};
+
+const prepareTreeData = (node, depth, maxDepth, weightAttribute) => {
```
aurel-l, line 27: Not a big fan of mutating the original data, in case it is shared with logic elsewhere.
I would keep the data in the component, untouched in _data, and prepare tree data would generate a new object, recursively cloned from the original data with all the required modifications done here, and assigned to a private _tree, _treeData, or _internalData, that would be used to draw.

packages/nightingale-sunburst/src/nightingale-sunburst.js
```diff
@@ -0,0 +1,358 @@
...
+    this["font-size"] = 10;
+    this.activeSegment = null;
+    this.holdSegment = false;
+    this.topOptions = superkingdoms;
+  }
+
+  prepareTree() {
+    if (!this._data) return;
+    prepareTreeData(this._data, 0, this["max-depth"], this["weight-attribute"]);
+    this.root = partition(this._data, this.side / 2, this["weight-attribute"]);
```
aurel-l, line 100: Ah, I see, so what is used to draw is root, so prepare data could return a clone of the data with the modifications instead of a mutated object, and that could be used here to pass to partition()

packages/nightingale-sunburst/src/nightingale-sunburst.js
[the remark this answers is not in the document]
aurel-l: OK, all good then 👌🏽

packages/nightingale-sunburst/src/nightingale-sunburst.js
gustavo-salazar: Cool, I'll work on a version of prepare data, that clones the original data.

packages/nightingale-sunburst/src/nightingale-sunburst.js
```diff
@@ -0,0 +1,358 @@
...
+      "mousemove",
+      (event) => {
+        if (!this.root || (this.activeSegment && this.holdSegment)) return;
+        this.selectNodeByPosition(
+          event.offsetX - this.side / 2,
+          event.offsetY - this.side / 2
+        );
+      }
+    );
+  }
```
aurel-l, line 349: Would put all of that in a connectedCallback, attaching the event listeners on this, and removing them on disconnectedCallback. It means the event listeners might need to be defined somewhere else (in the constructor?) and used in those 2 different lifecycle callbacks

aurel-l (approved):
All good apart from the 2 comments about:
 - trying to not mutate the original data
 - event listener registration
For me they're not really blocking though, so it depends on your workload, you might want to go ahead a merge and modify later, or wait to have these modifications in before merging and publishing

PR #178, opened 2021-11-11
[WIP] Nightingale new core storybook

dominikbruechner:
based on [#nightingalsen](https://github.com/ebi-webcomponents/nightingale/pull/176)

Adds storybook config with
- Navigation Component
- A11y
- webpack resolving aliases


gustavo-salazar:
Nevermind @dominikbruechner The actual issue is that I was using `displaystart` instead of `display-start` which is the new format of the attributes!

PR #179, opened 2021-11-11
2022 manager

xwatkins:
### Reference to issue
- Update test infrastructure
- Implement `nightingale-manager`

### Description of changes
- Added Jest configuration, as well as `ts-jest` to handle transpilation in the test context
- The config is at the root and ran in all packages thanks to lerna
- Re-implemented `nightingale-manager`, trying to simplify the code thanks to the Lit lifecycle methods:
   - Used custom converter on "attributes", renamed to "reflected-attributes"
   - Used state properties to hold list of registered HTMLElements and `propertyValues`

### Tests / Styleguide
 - [x] Tests pass
 - [ ] Styleguide


packages/nightingale-manager/README.md
```diff
@@ -0,0 +1,52 @@
...
+  new CustomEvent("change", {
+    detail: {
+      value: "New value",
+      type: "attr1",
+    },
+  })
+);
+```
+
+## API Reference
```
gustavo-salazar, line 36: I'm been a bit picky here, but maybe we should take this whole process of migrating into typescript to make sure the API documentation of each component is complete.

gustavo-salazar:
Hey @xwatkins 
I added a commit (9b59bda) to fix a race condition in the `withManager` and added `nightingale-manager` in the `dev/index.html` example.
Please have a look that I haven't break anything else

packages/nightingale-manager/tests/nightingale-manager.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  beforeEach(() => {
+    rendered = new NightingaleManager();
+    document.documentElement.appendChild(rendered);
+  });
+
+  afterEach(() => {
+    document.documentElement.removeChild(rendered);
+  });
+
+  test("It should propagate reflected attributes", () => {
```
gustavo-salazar, line 16: I think a better test would be to add 2 `innerElements` and make sure the atrributes get propagated to each other. Just to make sure is not just Lit doing it internally

packages/nightingale-navigation/src/nightingale-navigation.ts
```diff
@@ -91,16 +88,16 @@ class NightingaleNavigation extends withManager(
       ])
       .on("brush", ({ selection, transform }) => {
         if (selection && this.#x) {
-          this['display-start'] =
+          this["display-start"] =
```
gustavo-salazar, line 91: Sorry, I didn't notice my prettier is setup differently. We should add a prettier config to the project to make sure we avoid this sort of changes.

packages/nightingale-new-core/src/nightingale-base-element.ts
```diff
@@ -3,8 +3,8 @@ import { LitElement, html } from "lit";
 export type Constructor<T = {}> = new (...args: any[]) => T;
 
 class NightingaleElement extends LitElement {
-  render() {
-    return html` <p>NightingaleElement.</p> `;
+  createRenderRoot() {
```
gustavo-salazar, line 6: @xwatkins @aurel-l @dlrice @dominikbruechner 
FYI I'm adding this to the root `NightingaleElement` so it renders in the normal DOM and not in shadow DOM, as the latter causes some issues when working with d3.

gustavo-salazar (approved):
Have a look to the comments I put around.
Otherwise happy to merge!

PR #180, opened 2021-12-23
Tooltip refactor

xwatkins:
Note: not happy with the casting to `unknown` in `protvista-feature-adapter` but haven't found a better solution to override so far... needs discussion.

### Tests / Styleguide
 - [ ] Tests pass
 - [ ] Styleguide


gustavo-salazar:
Hey doing a bit of a clean up around, is this still happening, maybe @aurel-l @dlrice or @Swaathik can review if is completed? We don't use the tooltip component in InterPro, so I wouldn't know if after this changes it is doing what it suppose to.

PR #182, opened 2022-01-21
Interaction ts

xwatkins:
# Done
- Use Lit
- Move to TS
- Define data model
- Use `data-loader`
- Use `protvista-tooltip`
- Parse data, check with Jie for isoforms
- Display adjacency graph
- Update tooltip generation logic
- Update filter logic, use standard components

# TODO
## Graph
- on hover, highlight corresponding 2 protein names, and add lines (see current implementation)
- Hide redundant part of graph (?? UX check)
## Filters
~~subcellular component should be displayed as a tree (?? UX check)~~


packages/interaction-viewer/src/AdjacencyGraph.ts
```diff
@@ -0,0 +1,303 @@
...
+  };
+
+  const getIntactLink = (interactor1: string, interactor2: string): string => {
+    return `//www.ebi.ac.uk/intact/query/id:${interactor1} AND id:${interactor2}`;
+  };
+
+  const mouseover = (accession1: string, accession2: string) => {
+    // svg.classed("active-cell", true);
+
+    const group = svg.insert("g", ":first-child").attr("class", "active-group");
```
aurel-l, line 96:
```suggestion
    const group = svg.insert("g").attr("class", "active-group");
```
In order to have the line go over the grey triangle so that it's easier to see to what accession we are pointing

aurel-l:
Since the example data has changed, it would be good to replace the accession in the documentation markdown too

aurel-l (commented):
Some suggestions around the lit lifecycle

packages/interaction-viewer/src/interaction-viewer.ts
```diff
@@ -0,0 +1,159 @@
...
+  processedData?: ProcessedData;
+
+  @property()
+  filteredAccessions: string[] = [];
+
+  private handleFilterSelection(event: CustomEvent): void {
+    this.filteredAccessions = event.detail;
+  }
+
+  async connectedCallback(): Promise<void> {
```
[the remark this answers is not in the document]
xwatkins: I looked this up and couldn't find anything saying it couldn't be, even found a tutorial that did exactly this. I agree that it looks weird though.

packages/interaction-viewer/src/interaction-viewer.ts
```diff
@@ -0,0 +1,159 @@
...
+    this.addEventListener(FILTER_SELECT, this.handleFilterSelection);
+
+    if (!this.accession) {
+      return;
+    }
+    const response = await load(
+      `https://www.ebi.ac.uk/proteins/api/proteins/interaction/${this.accession}.json`
+    );
+    const data = response.payload as APIInteractionData[];
+    this.processedData = process(data);
```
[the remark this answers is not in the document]
xwatkins: The issue with having it in `updated` is that it keeps triggering itself by re-triggering updates when properties change. I think `updated` provides a Map of changed properties, so it should be possible to only trigger if this is the case... will have a look.

packages/interaction-viewer/src/interaction-viewer.ts
```diff
@@ -0,0 +1,174 @@
...
+import styles from "./styles";
+import { APIInteractionData } from "./data";
+// eslint-disable-next-line import/no-duplicates
+import InteractionTooltip from "./interaction-tooltip";
+// eslint-disable-next-line import/no-duplicates
+import { FILTER_SELECT } from "./interaction-filters";
+
+// Import additional components
+// eslint-disable-next-line import/no-duplicates
+import "./interaction-filters";
```
xwatkins, line 22: @aurel-l meant to ask your opinion about this - I wonder if it would be better to have all the WC imports in `index.ts`?

packages/interaction-viewer/src/interaction-viewer.ts
```diff
@@ -0,0 +1,159 @@
...
+};
+
+@customElement("interaction-viewer")
+export default class InteractionViewer extends LitElement {
+  private nodes: APIInteractionData[] = null;
+
+  @property()
+  accession: string;
+
+  @property()
```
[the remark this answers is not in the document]
aurel-l: See recommendations here https://lit.dev/docs/components/properties/#internal-reactive-state

packages/interaction-viewer/src/interaction-viewer.ts
aurel-l: Not sure. Originally it was to have the index.ts do the WC registration _and_ re-export, but we decided it was not really useful to expose an unregistered element and have the user register it, and you have the decorator doing the registration in the other file anyway. I think it's important that the entry point remains index.ts, because that's what the tooling will look for as an entry point and we don't want to change the tooling config too much across every package.

packages/interaction-viewer/src/interaction-viewer.ts
```diff
@@ -0,0 +1,177 @@
...
+    if (Array.from(changedProperties.keys()).includes("accession")) {
+      if (!this.accession) {
+        return;
+      }
+      const response = await load(
+        `https://www.ebi.ac.uk/proteins/api/proteins/interaction/${this.accession}.json`
+      );
+      const data = response.payload as APIInteractionData[];
+      this.processedData = process(data);
+    }
```
aurel-l, line 114: Issue with async is that we might start to get race conditions

```suggestion
  updated(changedProperties: Map<string, any>): Promise<void> {
    // Only run this if it's the accession that has changed
    if ('accession' in changedProperties) {
      if (!this.accession) {
        return;
      }
      // fetch then later on update the state
      load(
        `https://www.ebi.ac.uk/proteins/api/proteins/interaction/${this.accession}.json`
      ).then((response) => {
        // eventually update the state
        this.processedData = process(response.payload as APIInteractionData[]);
      });
      // but first remove any previous data from the state (effectively, we are in a loading state)
      this.processedData = null;
    }
```

aurel-l (approved):

Issue #202, opened 2022-07-28
Missing height setter in protvista-structure


radusuciu:
I'm attempting to use `ProtvistaStructure` in a vue3 project and when attempting to set the height attribute I get this error:

```
[Vue warn]: Failed setting prop "height" on <protvista-structure>: value 329px is invalid. TypeError: Cannot set property height of #<ProtvistaStructure> which has only a getter
```

Looking [at the source](https://github.com/ebi-webcomponents/nightingale/blob/master/packages/protvista-structure/src/protvista-structure.ts#L138-L140), I see that while there is a getter defined for `height`, there is no `setter`.

radusuciu:
Hi, I thought that this made it into the `3.6.1` release of this package, but peeking in the `dist` folder, it seems like it has not been included since there are no instances of `set height` and I still get the same error. Any chance the sources could've been updated but not built?

PR #204, opened 2022-07-29
Remove return type annotation on protvista-structure height setter

radusuciu:
This was an oversight on the original PR related to #202.


xwatkins (approved):
Well spotted

Issue #206, opened 2022-08-14
Add support for structure highlight across gaps in mapping


prihoda:
Problem 1: Highlighting a structure region is currently not supported across alignment gaps in the sequence-structure mapping.

For example, this PDB contains an insertion at position 41, so the mapping contains two items, one from uniprot position 1-40, the second from uniprot position 41-164: https://www.ebi.ac.uk/pdbe/api/mappings/uniprot/102L

When trying to highlight range 37-42, the current position mapping fails, because the start and end coordinates are not found in the same mapping item:

https://github.com/ebi-webcomponents/nightingale/blob/91baedb57d58d40d784fcf7acd7e65c14bbcde52/packages/protvista-structure/src/position-mapping.ts#L63-L73

Problem 2: When multiple chains are present, highlight is applied to all chains, even though it should only be applied to specific chains present in the mapping.

The molstar condition only checks the residue number, not the chain:

https://github.com/ebi-webcomponents/nightingale/blob/91baedb57d58d40d784fcf7acd7e65c14bbcde52/packages/protvista-structure/src/structure-viewer.ts#L199-L206

You can try this out by selecting structure `1BRC` from the table at https://ebi-webcomponents.github.io/nightingale/#/structure and clicking at residues in the smaller chain. Residues in the larger chain will also be highlighted.

gustavo-salazar:
@prihoda I'm guessing #207 solves this issue. Thanks for the contribution.
I'm closing this ticket, but feel free to reopen it if there is anything missing.

PR #207, opened 2022-08-14
Add support for mapping with gaps and different chains

prihoda:
### Reference to issue

See https://github.com/ebi-webcomponents/nightingale/issues/206

### Description of changes

- Add support for selecting a range across mapping regions (across gaps in mapping)
- Make sure that only chains present in mapping are highlighted

See comments in code for more context

### DIFF
--- a/packages/protvista-structure/src/position-mapping.ts
+++ b/packages/protvista-structure/src/position-mapping.ts
@@ -53,7 +53,9 @@ const translatePositions = (
     throw new PositionMappingError("Invalid start, end coordinates");
   }
   // return a translation separately for each chain (if it exists)
-  return Object.entries(groupBy(mappings, (mapping) => mapping.chain_id))
+  const translations = Object.entries(
+    groupBy(mappings, (mapping) => mapping.chain_id)
+  )
     .map(([chain_id, chainMappings]) => {
       let startMapping = null;
       let endMapping = null;
@@ -82,9 +84,9 @@ const translatePositions = (
         }
       }
       if (startMapping === null || endMapping === null) {
-        throw new PositionMappingError(
-          "Start or end coordinate outside of mapping range"
-        );
+        // If we didn't find a mapping for this chain,
+        // return null, it will be filtered out later
+        return null;
       }
       const direction = mappingDirection === "UP_PDB" ? 1 : -1;
       return {
@@ -100,6 +102,14 @@ const translatePositions = (
       };
     })
     .filter(Boolean);
+
+  if (!translations.length) {
+    throw new PositionMappingError(
+      "Start or end coordinate outside of mapping range"
+    );
+  }
+
+  return translations;
 };
 
 export default translatePositions;
--- a/packages/protvista-structure/src/position-mapping.ts
+++ b/packages/protvista-structure/src/position-mapping.ts
@@ -1,3 +1,5 @@
+import groupBy from "lodash-es/groupBy";
+
 export type Direction = "UP_PDB" | "PDB_UP";
 
 /* eslint-disable camelcase */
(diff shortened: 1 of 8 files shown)


packages/protvista-structure/src/position-mapping.ts
```diff
@@ -41,51 +43,63 @@ const translatePositions = (
   end: number;
   entity: number;
   chain: string;
-} | null => {
+}[] => {
```
prihoda, line 46: we return a list of mapped results, one per each chain present in the mapping

packages/protvista-structure/src/position-mapping.ts
```diff
@@ -41,51 +43,73 @@ const translatePositions = (
...
-        ? mapping.start.residue_number - mapping.unp_start
-        : mapping.unp_start - mapping.start.residue_number;
-    return {
-      entity: mapping.entity_id,
-      chain: mapping.chain_id,
-      start: start + offset,
-      end: end + offset,
-    };
+  // return a translation separately for each chain (if it exists)
+  const translations = Object.entries(
```
prihoda, line 56: Group the list of mapping regions by their chain, then iterate over the chain mapping regions. If a translation for the given chain is found, it will be added to the list. If it's not found, we return null which will be filtered out. If a mapping is not found for any chain, we throw an exception, same as before.

packages/protvista-structure/src/protvista-structure.ts
```diff
@@ -369,22 +384,21 @@ class ProtvistaStructure extends HTMLElement implements NightingaleElement {
...
+        if (this.isAF()) {
+          return {
             start,
             end,
-            this._selectedMolecule.mappings,
-            "UP_PDB"
-          );
-        })
-        .filter(Boolean);
+            chain: "A",
```
prihoda, line 392: The AlphaFold chain is always called A.

prihoda:
@gustavo-salazar can you please check this PR? Thanks

gustavo-salazar:
Hey @prihoda sorry I only have the time to look at this now.
I have a couple of questions:
1. Is it possible to have a protein split into 2 chains? For instance: protein A with length 50, which has structure X with chain A covering 1 to 20 and structure B from 21 to 50. I really don't know if this is possible in the world of PDB structures, but if that's the case, I'm unsure your algorithm will cover that situation.
2. What if the region to select covers 3 mappings? If I understood correctly you are ignoring the middle ones, and it calculates the translation only considering the start and end mappings. This actually seems sensible to me, but just want to make sure you have thought of this scenario and if there are any edge case you need to take into account here.

prihoda:
Hi @gustavo-salazar 

1. Not sure if this can happen for uniprot mappings, but you're right that this is not handled. Currently we expect that the start and end of the range is in the same chain, we are just using the start chain:
 
https://github.com/ebi-webcomponents/nightingale/blob/112100bf4531295920df40ecc3cb2599623afadd/packages/protvista-structure/src/position-mapping.ts#L92-L102

Highlighting across chains would probably be difficult to implement, so I suggest we can just detect this and return an error in such a special case. What do you think? 

2. The mapping is just used to determine the start and end position in the chain. So as long as the mappings belong to the same chain, it doesn't matter which mapping we find the start/end coordinate, the output is just the chain id, the start coordinate in the chain and the end coordinate in the chain.

gustavo-salazar:
1. Yeah we can do that. And if we end up finding that the case actually happens we can work on that at that point.
2. Cool, that's what I thought, thanks for explaining.

I'm going to merge this PR and will republish the component later today

prihoda:
Awesome. Since you've merged this, should I create a second PR with the fix to detect that 1. scenario?

gustavo-salazar:
🙈 Good point!

I think that if you push again in the same branch, we can reopen the PR, but I'm not sure. we can give that a try otherwise, yes please open a 2nd PR

PR #209, opened 2022-09-19
Add test case for structure mapping between chains

prihoda:
Turns out that trying to select a region across two chains already shows an error "Start or end coordinate outside of mapping range". This is because the mappings are processed separately for each chain:

https://github.com/ebi-webcomponents/nightingale/blob/master/packages/protvista-structure/src/position-mapping.ts#L55-L61

I added a test case demonstrating this behavior.

### DIFF
--- a/packages/protvista-structure/src/__test__/__mocks__/mapping-test-cases.js
+++ b/packages/protvista-structure/src/__test__/__mocks__/mapping-test-cases.js
@@ -43,6 +43,19 @@ const testCases = [
   {
     entry: "1AAP",
     pdb: {
+      start: 10,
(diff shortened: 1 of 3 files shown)


prihoda:
@gustavo-salazar please check

prihoda:
@gustavo-salazar please check

gustavo-salazar (approved):

gustavo-salazar:
Hey @prihoda I have published a new version of `protvista-structure` that includes the mapping feature: https://www.npmjs.com/package/protvista-structure

Please have a look and let me know if there is any issue

PR #214, opened 2022-10-19
Nightingale interpro track

gustavo-salazar:
Adding labels to the Interpro tracks


packages/nightingale-interpro-track/src/nightingale-interpro-track.ts
```diff
@@ -258,6 +302,51 @@ class NightingaleInterproTrack extends NightingaleTrack {
...
+
+        if (firstPath) {
+          return contrastingColor(getColor(firstPath, "fill"));
+        }
+        return null;
+      })
+      .attr(
+        "font-size",
+        (f) => (this.layoutObj?.getFeatureHeight(f.feature as Feature) || 2) - 2
+      )
```
matthiasblum, line 341: Makes the child labels _very_ small. Is the `-2` really necessary?

packages/nightingale-interpro-track/src/nightingale-interpro-track.ts
```diff
@@ -258,6 +302,51 @@ class NightingaleInterproTrack extends NightingaleTrack {
...
+    return datum?.feature?.accession || null;
+  }
+
+  private refreshLabels(base: LabelGroup) {
+    base
+      .attr("x", (f) => {
+        const start = f.start || 0;
+        const end = f.end || 0;
+        return this.getXFromSeqPosition(start + (end - start) / 2 || 1);
+      })
```
matthiasblum, line 321: This causes the label to shift out of the viewport when zooming.

Something like this could help keeping the label visible:

```js
const displayStart = this["display-start"] || 0;
const displayEnd = this["display-end"] || this.length;
const start = Math.max(f.start || 0, (displayStart as number));
const end = Math.min(f.end || 0, (displayEnd as number));
return this.getXFromSeqPosition(start + (end - start) / 2 || 1);
```

The `wrap()` function would have to be modified.

matthiasblum (approved):
Looks good. I made a few comments, but nothing preventing merging the PR.

I am not sure if there is a policy not to tamper with style in `nightingale` and let projects using it dealing with style themselves, but I would suggest to:

- use `cursor: default;` on `text` elements
- use `cursor: pointer;` on `g.location-group` elements that can be expanded/collapsed.

packages/nightingale-interpro-track/src/nightingale-interpro-track.ts
gustavo-salazar: I thought of it as a bit of padding, 1 pixel up and 1 down.
I'm making it that the children don't have padding

gustavo-salazar:
@matthiasblum Ok changes done, please check again.

matthiasblum (approved):
Looks good to me.

PR #225, opened 2022-11-10
nightingale-structure

aurel-l:
### Description of changes

Uses Molstar and reimplements a structure viewer with the new 2022 logic

Missing:
 - [ ] interaction in the structure (hover, click) should emit events understandable by nightingale-manager
 - [ ] AlphaFold colouring

### Tests / Styleguide
 - [ ] Tests pass
 - [ ] Styleguide


prihoda:
FYI latest Mol* does AlphaFold pLDDT coloring automatically, you can try it here if you drag and drop the cif file  https://molstar.org/viewer/

gustavo-salazar:
Remember to merge into `main` instead of `2022`

stories/08.Structure/NightingaleStructure.stories.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const Structure = Template.bind({});
+Structure.args = {
+  "protein-accession": "P05067",
+  "structure-id": "1AAP",
+  highlight: "290:315",
+  "highlight-color": DEFAULT_HIGHLIGHT_COLOR,
+  "--custom-structure-height": "480px",
+};
+Structure.play = async () => {
```
gustavo-salazar, line 37: you can delete all of this as there is nothing to setup after is loaded

stories/08.Structure/NightingaleStructure.stories.ts
```diff
@@ -0,0 +1,43 @@
...
+  await customElements.whenDefined("nightingale-structure");
+  const track = document.getElementById("viewer");
+  // if (track) {
+  //   (track as any).data = tinyData;
+  // }
+};
```
gustavo-salazar, line 43: Maybe add another story that loads alphafold structures

gustavo-salazar (approved):
Is there a way to specify a chain when highlighting? I guess it makes sense to select multiple chains when it is the same protein, but if a structure has chains from different proteins, I think it is necessary to be able to indicate which chain to select.

I don't understand the need of the `protein-accession` attribute, or is it related with what I'm asking above about chains?

In any case I think we could release this version and treat everything as issues to fix later.

PR #226, opened 2022-11-10
Fix for issue with disappearing D3 axis tick in sequences

Two-Kay:
A rounding error for D3 Axes in [nightingale-sequence](https://github.com/ebi-webcomponents/nightingale/blob/8d22e6c4027a6b9c693c5683ca386d3c309494aa/packages/nightingale-sequence/src/nightingale-sequence.ts#L195-L200) caused the starting tick to disappear if width/length was modified to certain values.

[The xScale for the axis object is copied and its domain is rounded as decimal numbers are irrelevant for D3 Axes.](https://github.com/Two-Kay/nightingale/blob/267a2541507d1ef917a8f5ce3df9b968e06450db/packages/nightingale-sequence/src/nightingale-sequence.ts#L196-L198)

### DIFF
--- a/packages/nightingale-sequence/src/nightingale-sequence.ts
+++ b/packages/nightingale-sequence/src/nightingale-sequence.ts
@@ -193,7 +193,9 @@ class NightingaleSequence extends withManager(
 
       // only add axis if there is room
       if (this.height > (this.chWidth || 0) && this.xScale) {
-        const xAxis = axisBottom(this.xScale)
+        let roundScale = this.xScale;
+        roundScale.domain(roundScale.domain().map((i:number) => Math.round(i)));
+        const xAxis = axisBottom(roundScale)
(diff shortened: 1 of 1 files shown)


packages/nightingale-sequence/src/nightingale-sequence.ts
```diff
@@ -193,7 +187,12 @@ class NightingaleSequence extends withManager(
 
       // only add axis if there is room
       if (this.height > (this.chWidth || 0) && this.xScale) {
-        const xAxis = axisBottom(this.xScale)
+        // Copying the scale, and rounding its domain values to avoid bug releated with floats
+        const roundScale = this.xScale.copy();
```
gustavo-salazar, line 191: Hey @Two-Kay 
I added the `.copy()` here to make sure changes in `roundScale` won't affect `xScale`

PR #228, opened 2022-11-15
Handle async init of molstar

prihoda:
In our app I discovered that there are race conditions with configuring the structure viewer.

In StructureViewer constructor, this line sometimes fails because the molstar plugin has not been initialized yet:
https://github.com/ebi-webcomponents/nightingale/blob/v3/packages/protvista-structure/src/structure-viewer.ts#L139-L140

This PR solves this by wrapping that piece of code into a callback that's executed when molstar canvas is initialized.

### DIFF
--- a/packages/protvista-structure/src/structure-viewer.ts
+++ b/packages/protvista-structure/src/structure-viewer.ts
@@ -121,33 +121,35 @@ class StructureViewer {
         onHighlightClick([{ position: sequencePosition, chain: chain }]);
       }
     });
-    PluginCommands.Canvas3D.SetSettings(this.plugin, {
-      settings: (props) => {
-        // eslint-disable-next-line no-param-reassign
-        props.renderer.backgroundColor = Color(0xffffff);
-      },
+    const that = this;
+    this.plugin.behaviors.canvas3d.initialized.subscribe((v) => {
+      if (v) {
+        PluginCommands.Canvas3D.SetSettings(that.plugin, {
+          settings: (props) => {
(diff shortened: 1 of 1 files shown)


packages/protvista-structure/src/structure-viewer.ts
```diff
@@ -121,33 +121,35 @@ class StructureViewer {
         onHighlightClick([{ position: sequencePosition, chain: chain }]);
       }
     });
-    PluginCommands.Canvas3D.SetSettings(this.plugin, {
-      settings: (props) => {
-        // eslint-disable-next-line no-param-reassign
-        props.renderer.backgroundColor = Color(0xffffff);
-      },
+    const that = this;
```
gustavo-salazar, line 124: I don't think this is necessary. `this` should work because you are using arrow functions. or am I missing something?

prihoda: You are right, it works, I updated the PR 👍

gustavo-salazar (approved):

gustavo-salazar:
Hey @prihoda, thanks for the PR. I just merged and republished the component as [3.8.2](https://www.npmjs.com/package/protvista-structure)
please check.

gustavo-salazar:
Hey @prihoda Sorry to kidnap this PR, but I didn't know any other way to contact you.

We are writing a paper about nightingale. And we want to list projects that are using our components. Do you mind if we include yours in the list? if you are Ok with it do you mind sending me a link to a place where you are using it?
Thanks. Gustavo

PR #229, opened 2022-11-16
Revert rounding of xAxis

Two-Kay:
While the rounding fixes the missing ticks on the sole sequence component without causing any further issues while isolated, I did notice it causes issues in the manager component since it affects the exact position of a tick there as the sequence doesn't have to start or end at a rounded number.
Only in its isolated state when using an integer starting point, the floating point error will become noticeable.

### DIFF
--- a/packages/nightingale-sequence/src/nightingale-sequence.ts
+++ b/packages/nightingale-sequence/src/nightingale-sequence.ts
@@ -187,12 +187,7 @@ class NightingaleSequence extends withManager(
 
       // only add axis if there is room
       if (this.height > (this.chWidth || 0) && this.xScale) {
-        // Copying the scale, and rounding its domain values to avoid bug releated with floats
-        const roundScale = this.xScale.copy();
-        roundScale.domain(
-          roundScale.domain().map((i: number) => Math.round(i))
-        );
-        const xAxis = axisBottom(roundScale)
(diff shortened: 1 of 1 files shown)


gustavo-salazar:
Hey @Two-Kay 
I haven't been able to replicate the error you mentioned here. Do you mind trying using what is in branch `2022`, and doing a `yarn clean`, then `yarn build` and then testing the `yarn storybook`?
I introduced a change in the core code related to the height of the component, but it could have some side-effects so I'm wondering if could be related.

Two-Kay:
> Hey @Two-Kay I haven't been able to replicate the error you mentioned here. Do you mind trying using what is in branch `2022`, and doing a `yarn clean`, then `yarn build` and then testing the `yarn storybook`? I introduced a change in the core code related to the height of the component, but it could have some side-effects so I'm wondering if could be related.

Hi @gustavo-salazar 
I'm on branch 2022 and still encounter the same issue, where you can see on the [manager](https://github.com/ebi-webcomponents/nightingale/blob/420f81411abfddbd79fcd0b1009b374d4b9b846d/packages/nightingale-manager/src/nightingale-manager.ts) package, that the xAxis ticks on the sequence component are now off-center while the rounding is active:

Off centered xAxis ticks

PR #238, opened 2023-01-30
Nightingale links

gustavo-salazar:
Migration: `protvista-links => nightingale-links`

**Note:**  Review this after #237


packages/nightingale-links/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Attributes
+
+#### `min-distance: number (default: 0)`
+
+It filters out the links between bases that are closer than the given minimum distance.
+
+#### `min-probability: number (default: 0.7)`
+
+It filters out the links that have lessprobability than the given one.
```
matthiasblum, line 29:
```suggestion
It filters out the links that have less probability than the given one.
```

packages/nightingale-links/src/nightingale-links.ts
```diff
@@ -0,0 +1,231 @@
...
+          this._dispatchSelectNode(d);
+          this.refresh();
+        })
+        .on("mouseout", () => {
+          if (!this.#contacts || this.#contacts?.isHold) return;
+          this.#contacts.selected = undefined;
+          this.dispatchEvent(getHighlightEvent("mouseout", this));
+          this.refresh();
+        })
+        .on("click", (d: number) => {
```
matthiasblum, line 153:
```suggestion
        .on("click", (_: Event, d: number) => {
```

Otherwise `d` is the click event, and when passed to `_dispatchSelectNode`, `this.#contacts.contacts[d]` is `undefined`.

matthiasblum (changes requested):
Spotted a small bug that breaks the click event.

I also encountered an unexpected behavior, but I am not sure what is the cause. When clicking and holding a circle, then slowing dragging the cursor outside of the circle, the view zooms in:

output

Finally, a few suggestions:
- when selecting a point, it is highlighted (bigger size + green border). I'd highlight the linked points as well (maybe just bigger size, so just the clicked point has a green border)
- when a point is selected, clicking another point unselect the first point and highlight the second, but I think it should also select the second point
- would it be possible to unselect a point, when clicking anywhere _outside_ a point?

gustavo-salazar:
OK the suggestions and requested changes are in the branch [fixesPreRelease](https://github.com/ebi-webcomponents/nightingale/tree/fixesPreRelease). I'll merge this after #237

PR #241, opened 2023-02-06
Nightingale heatmap

gustavo-salazar:
migration of `nightingale-heatmap`

It also has dependencies updates and some bug fixes when the components were tested out of the storybook.

**Note:** Check after #239


matthiasblum (approved):
Looks good to me. I have two minor suggestions, though.

---

With `margin-top=0` (default), the top tick of the y-axis is truncated:

image

This tick could be stylized by adding the following code:

```ts
    svg
        .select<SVGGElement>('g.y-axis g.tick')
        .attr("dominant-baseline", "hanging");
```

after 

https://github.com/ebi-webcomponents/nightingale/blob/4735e8f43424806b7f2a9a06c8553c0d943915cb/packages/nightingale-heatmap/src/nightingale-heatmap.ts#L231-L236

Alternatively, `nightingale-heatmap` could override the default margin values from `withMargin` to:

- `margin-top=5`
- `margin-right=5`
- `margin-bottom=50`
- `margin-left=50`

---

I like how the hovered point is highlighted, but the default border color (from `withHighlight`) is a pale yellow that may work with darker colors, but isn't really visible if the heatmap uses lighter colors.

The default value of `highlight-color` could be overridden to something darker (e.g. `#9e9e9e`)

gustavo-salazar:
> The default value of `highlight-color` could be overridden to something darker (e.g. `#9e9e9e`)

Good point. I'm going to put `#fc1e1e` as the default color as it is triad complementary to the `#23368a` and `#fff81f` so it should offer good contrast to both. I'm also increasing the width of the border on that circle, making it more prominent.

These changes are in #245 .

gustavo-salazar:
> With `margin-top=0` (default), the top tick of the y-axis is truncated:

I went for the `dominant-baseline` suggestion. Change in #245

Issue #227, opened 2022-11-14
Standardize color VS colour


prihoda:
Hi all, since you are creating a new major version, I would suggest standardizing using `color` vs `colour`:

See usages of `colour`: https://github.com/ebi-webcomponents/nightingale/search?q=colour

PR #242, opened 2023-02-09
Colour => Color

gustavo-salazar:
This PR normalised the use of the word `color` instead of `colour` in the codebase as suggested in #227 

Most of the changes happened in the component `nightingale-colored-sequence`.

**Note:** Review after #241


matthiasblum (approved):
All good.

PR #243, opened 2023-02-16
README: add build to locally showcase application

dlrice:
Had to do this to get the storybook working for me otherwise I encountered TS errors.


gustavo-salazar (approved):
Yeah you are right, the build is necessary. thanks for adding it in the readme

PR #252, opened 2023-04-05
Getting started & creating a component documentation

dlrice:
- Created getting started doc which uses Sandpack to show demo code.
- Created creating a component


stories/002.CreatingAComponent.stories.mdx
```diff
@@ -0,0 +1,185 @@
...
+      export default NightingaleHello;
+      ```
+
+      Here are some notes of this code:
+
+      - The library used to build the web components in Nightingale is [Lit](https://lit.dev/docs/).
+      - The `@customElement` decorator is used for defining the custom element following the web component standard of defining and registering custom elements.
+      - The component rendering is defined by the method `render()`.
+
+   2. The above example inherits from [LitElement](https://lit.dev/docs/api/LitElement/), while there is a base element for any Nightingale component named `NightingaleElement`.
```
gustavo-salazar, line 102: I think is worth splitting this in 2 points:
2. Use `NightingaleElement` instead of `LitElement`...
3. Use of mixins, maybe adding one at a time.

gustavo-salazar (changes requested):
Nice job!
I added a few minor suggestions/comments in both tutorials, but I think in general they are what we are looking for.

dlrice:
@gustavo-salazar addressed all of the getting started comments.  In the end I had a refactor to try make it flow a bit better.

gustavo-salazar (approved):

PR #253, opened 2023-04-11
Uniprot showcase example

dlrice:


gustavo-salazar:
This is great Dan! For the purpose of the paper, we just need to add the structure.
Do you have the time to make it more in the shape of a tutorial?
I'm thinking of something like:
1. take the result from getting starting tutorial
2. Fetch data from Uniprot and set it in the sequence component
3. Add navigation and manager to see the zooming and panning work.
4.  Add tracks that use `featuresData`
5. Add the variation track
6. Add the structure component

stories/003.LoadingUniProtData.stories.mdx
```diff
@@ -0,0 +1,287 @@
...
+        <td>Domain</td>
+        <td>
+          <nightingale-track
+            id="domain"
+            min-width="800"
+            height="15"
+            length="770"
+            display-start="1"
+            display-end="770"
+            margin-color="aliceblue"
```
gustavo-salazar, line 179: Add `highlight-event="onmouseover"` here and in other track

dlrice:
Addressed all comments with the exception of adding the structure which depends on https://github.com/ebi-webcomponents/nightingale/pull/225

gustavo-salazar (approved):
This is really cool! thanks!
I'm going to merge this, and @aurel-l can either open another PR, or added on the one he is migrating the structure

PR #256, opened 2023-05-25
Binding site tooltip and PTM changes

Swaathik:
https://www.ebi.ac.uk/panda/jira/browse/TRM-29602

The PTMs tooltip has to be merged (dbReferences in particular) and show only one peptide feature in place of many only when the PTM site is present in the same position in the same length peptide.


packages/protvista-proteomics-adapter/src/ProtvistaProteomicsAdapter.js
```diff
@@ -1,20 +1,83 @@
...
+    const ptmMap = {};
+    data.features.forEach((feature) => {
+      let ft = `${feature.begin}-${feature.end}`;
+      if (feature.ptms) {
+        feature.ptms.forEach((ptm) => {
+          ft += `-${ptm.position}`;
+        });
+        if (ptmMap[ft]) {
+          ptmMap[ft] = [...ptmMap[ft], feature];
+          ptmMap[ft].push(feature);
```
dlrice, line 30: `feature` seems to be added twice here

packages/protvista-proteomics-adapter/src/ProtvistaProteomicsAdapter.js
```diff
@@ -1,20 +1,83 @@
...
+        } else {
+          ptmMap[ft] = [feature];
+        }
```
dlrice, line 33:
```suggestion
        ptmMap[ft] = ft in ptmMap ? [...ptmMap[ft], feature] : [feature]
```
I find ternaries quite useful in these situations

packages/protvista-proteomics-adapter/src/ProtvistaProteomicsAdapter.js
```diff
@@ -1,20 +1,83 @@
...
+      }
     });
+
+    if (Object.keys(ptmMap).length) {
+      adaptedData = Object.values(ptmMap).map((value) => {
```
dlrice, line 38: Rather than `value` what do you think about something like `features`? Could make it a little easier to read in the future.

packages/protvista-proteomics-adapter/src/ProtvistaProteomicsAdapter.js
```diff
@@ -1,20 +1,83 @@
...
+        // ONly the dbReferences have to be merged as the rest is all the same
+        const mergedDbReferences = [];
+        value.forEach((feature) => {
+          feature.ptms.forEach((ptm) => {
+            ptm.dbReferences.forEach((dbReference) => {
+              const refExists = mergedDbReferences.find(
```
dlrice, line 44: If `mergedDbReferences` was a map with `dbReference.id` as the key the `find` could be avoided.

packages/protvista-proteomics-adapter/src/ProtvistaProteomicsAdapter.js
```diff
@@ -1,20 +1,83 @@
...
+          unique: value[0].unique,
+          ptms: value[0].ptms.map((ptm) => ({
+            name: ptm.name,
+            position: ptm.position,
+            sources: ptm.sources,
+            dbReferences: mergedDbReferences,
+          })),
+        };
+
+        return Object.assign(
```
dlrice, line 70: Out of curiosity was there a reason to avoid spreading?

dlrice (approved):

PR #259, opened 2023-06-11
Color by pLDDT using Molstar extension (DRAFT)

prihoda:
Proof of concept of pLDDT coloring directly using Molstar extension

This is just a draft for now. Temporarily disabled other stories in the storyboard due to typescript errors (at least on my end).


packages/nightingale-structure/src/structure-viewer.ts
```diff
@@ -32,17 +45,47 @@ const viewerOptions = {
   pdbProvider: "pdbe",
   viewportShowControls: PluginConfig.Viewport.ShowControls.defaultValue,
   viewportShowSettings: PluginConfig.Viewport.ShowSettings.defaultValue,
+  extensions: ObjectKeys(Extensions),
 };
 
+const ViewerAutoPreset = StructureRepresentationPresetProvider({
```
prihoda, line 51: This is also declared in `molstar/lib/apps/viewer/app.ts` but I assume we should not import from there but rather redeclare this class here https://github.com/molstar/molstar/blob/v3.34.0/src/apps/viewer/app.ts#L484-L510

prihoda:
@aurel-l @gustavo-salazar please take a look at the demo. We would need to make sure the initialization is done correctly and that it works consistently, but it works and only minimal code is needed :)

prihoda:
image

PR #265, opened 2023-07-12
Reflect properties

gustavo-salazar:
Changing some of the mixins to make the properties reflect as attributes.

Adding a validation that if the `length` is changed, then the maximum value of `display-end` should be the new `length`.

Updating test/snapshots to reflect changes above

Dependencies update, which required changes on some config files.

Sorry, running the new version of prettier changed a lot of files, so this PR has lots of minor code style changes.

*Side feature* The nightringale-saver component now accepts another attribute `scale-factor` which allows to scale the canvas before taking the picture, and might help improving the quality of the snapshot


matthiasblum (approved):
Looks good to me.

When reviewing the new `scale-factor` attribute for the `nightringale-saver` component, I spotted a small bug that occurs when enabling the `debug` option. I suggested a fix (+ other minor changes) in PR #267. Feel free to merge it in this branch if the fix is good enough.

PR #267, opened 2023-07-21
Minor fix in nightingale-saver

matthiasblum:
The size of the canvas (on which the image is drawn before being downloaded) is defined as such:

https://github.com/ebi-webcomponents/nightingale/blob/d77b9f52af3e3182a5bd92023ee06135a46f2452/packages/nightingale-saver/src/nightingale-saver.ts#L52

However, when `debug` is `true`, the canvas is added to `element`:

https://github.com/ebi-webcomponents/nightingale/blob/d77b9f52af3e3182a5bd92023ee06135a46f2452/packages/nightingale-saver/src/nightingale-saver.ts#L66

This results in `element` being higher and higher, and the downloaded image as well.

This PR fixes this by storing the initial height/width of `element` in private features.

Three other changes:

- rename `debug` with `preview`
- add the `preview` option to the `nightingale-saver`'s story in storybook
- when `preview` is `true`, only one canvas is added to the component. Any subsequent preview deletes the previous canvas.

### DIFF
--- a/packages/nightingale-saver/src/nightingale-saver.ts
+++ b/packages/nightingale-saver/src/nightingale-saver.ts
@@ -29,9 +29,6 @@ class NightingaleSaver extends NightingaleElement {
   @property({ type: Boolean })
   preview?: boolean = false;
 
-  #height: number = -1;
-  #width: number = -1;
-
   preSave?: () => void = undefined;
   postSave?: () => void = undefined;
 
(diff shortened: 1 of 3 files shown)


packages/nightingale-saver/src/nightingale-saver.ts
```diff
@@ -49,11 +52,17 @@ class NightingaleSaver extends NightingaleElement {
     if (typeof this.preSave === "function") {
       this.preSave();
     }
-    const { width, height } = element.getBoundingClientRect();
+    if (this.#height < 0 && this.#width < 0) {
```
gustavo-salazar, line 55: I think this doesn't consider the case where you have taken a snapshot, and then the element's size changes (e.g. when the tracks are responsive, and the window has been changed).

What if instead of saving previous dimensions, we make sure the `canvas.preview` is removed before getting the size?

gustavo-salazar (approved):

PR #270, opened 2024-02-08
MSA features

dlrice:
- Expose the setting of features for nightingale-msa
- Add story showing features and an event listener for clicks


gustavo-salazar (approved):
Looks good to me!!

Thanks Daniel, Feel free to merge.

My only comment is probably out of the scope of the PR: but just noticed that the basic MSA story is not working: It flashes on the screen and then disappear. If you can have a quick look and fix it, it would be great, otherwise, I'll open an issue and try to fix it whenever I have some time.

PR #286, opened 2024-05-20
Conservation inline worker

dlrice:
- Replace normal web worker with an inline web worker.
- Retained most of the code as normal TS but now `message` event listener is now a string.
- Removed the web worker rollup related package and config as now no longer used.
- Works in MSA storybook.
- Works when linked to uniprot codebase (using yalc).


gustavo-salazar:
Hey Daniel,

I haven't checked the PR yet, but besides the tests you mentioned in the description, I think it's important to test this directly from a published package. I usually publish a canary version and test using codepen ([example](https://codepen.io/gustavo-salazar/pen/ZEmWbLp)), but it could be just a plain HTML that takes the package from CDN.

dlrice:
[Here's a code pen with alpha version](https://codepen.io/Daniel-Rice/pen/jOoqZrY)

gustavo-salazar:
> [Here's a code pen with alpha version](https://codepen.io/Daniel-Rice/pen/jOoqZrY)

FYI, I did a quick test on the InterPro website with this alpha release and it seems to be working OK!

packages/nightingale-msa/src/workers/conservation-inline-worker.ts
```diff
@@ -40,16 +39,16 @@ export const calculateConservation = (
...
-
   return conservation;
 };
 
-const onmessage = function (e: MessageEvent) {
+const conservationInlineWorkerString = `
+self.addEventListener('message', (e) => {
   if (self.previous !== e.data) {
-    calculateConservation(e.data.sequences, e.data.sampleSize, true);
+    const f = ${calculateConservation.toString()};
```
gustavo-salazar, line 48: 🤯
wow, I'm really amazed by this trick!

gustavo-salazar (approved):
I like it!! thanks!
My only suggestion is that it would be good to include the link to inline-workers: https://web.dev/articles/workers-basics#inline_workers as a comment in the worker file as a reference of what is happening there. 

And maybe a line on how magically that line of `.toString()` gets typescript to transpile the function, so we could get most of the code still in TS.

packages/nightingale-msa/src/workers/conservation-inline-worker.ts
dlrice: I have to agree it was surprising when it worked!

PR #289, opened 2024-07-09
Feat/nightingale tooltip

justintyf01:
I am using this nightingale library for antibody sequence visualisation and it was missing a tooltip component. I have implemented this component, feel free to use it if deemed useful.

- Edited Feature type to include "description" and "evidence" instead of "tooltipContent".
- Main CSS styling for tooltip is in nightingale-tooltip
- Implemented a handleTrackClick in nightingale-manager that is called by the existing "change" event listener when a track is clicked
- Implemented new document event listener to handle closing of tooltip and de-highlighting when user clicks elsewhere on the screen

Thank you.

### DIFF
--- a/packages/nightingale-manager/src/nightingale-manager.ts
+++ b/packages/nightingale-manager/src/nightingale-manager.ts
@@ -3,155 +3,196 @@ import NightingaleElement from "@nightingale-elements/nightingale-new-core";
 
 @customElement("nightingale-manager")
 class NightingaleManager extends NightingaleElement {
-  @property({
-    converter: {
-      fromAttribute: (value): Map<string, null> | null => {
-        if (!value) {
-          return null;
-        }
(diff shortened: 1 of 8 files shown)


gustavo-salazar:
Dear @justintyf01 
Thank you so much for opening this PR, and sorry for taking this long to reply. 
And yeah, a tooltip is something other users have requested. So, I'll happily include a component that offers that into Nightingale.
I will check this PR soon but I'll ask you 2 things before:
1. The formatting on your side uses 4 spaces, and maybe other options that are making the read of your changes a harder than needed. I have run `npm run format` on `main`, so please do the same on your branch, after merging the latest changes from `main`. Hopefully. this will make the PR to only show the important changes.
2. Please create a story in `/stories` so we can see an example of the tooltip in the demo app.

PR #290, opened 2024-07-09
Nightingale sequence heatmap update to v1

marcelo-q:
Hello. The PDBe heatmap-component has been recently updated to it’s first stable version (1.0) and I’ve been asked to update the Nightingale Sequence Heatmap component to reflect this.

These are mostly minor fixes that do not affect what was previously discussed and adjusted in the first PR.

Perhaps the biggest change is the introduction of an attributed named "hm-highlight-width" to control the border width of the heatmap highlight.

Please let me know if there is anything requiring adjustment and I will be glad to reply and fix it

### DIFF
--- a/dev/index.html
+++ b/dev/index.html
@@ -101,7 +101,7 @@
         }
       });
       customElements.whenDefined("nightingale-sequence-heatmap").then(() => {
-        setTimeout( async() => {
+        setTimeout(async () => {
           const xDomain = [
             1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19,
             20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36,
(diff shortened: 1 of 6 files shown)


dev/index.html
```diff
@@ -185,6 +190,7 @@ <h1>Manager</h1>
           height="500"
           display-start="5"
           display-end="25"
+          hm-highlight-width="0"
```
gustavo-salazar, line 193: If I understand correctly, the only difference in not adding this attribute is the border of the highlighted area defaults to 1, isn't it?

dev/index.html
```diff
@@ -124,6 +124,11 @@
           document
             .getElementById("heatmapHotmap")
             .setHeatmapData(xDomain, yDomain, data);
+
+          // await document.getElementById("heatmapHotmap").updateComplete.then(() => {
```
gustavo-salazar, line 128: Delete this line if the are not going to be in use

gustavo-salazar (approved):
Hey there, 
Thanks for updating it to the latest version.
I put a couple of minor comments, but otherwise I'm happy to merge this. 
The question about the new attribute is mostly about deciding if this could be release as a minor version. i.e. not breaking changes

dev/index.html
marcelo-q: Done

dev/index.html
marcelo-q: Yes. This was added so the Nightingale Sequence Heatmap matches the highlighted style of Nightingale better after the PDBe heatmap-component fixes

gustavo-salazar:
Hey thanks for the changes. before I merge, do you mind solving the conflict with `main`, I think it's because I ran prettier on the `main` branch after you have branch out

PR #307, opened 2024-10-15
NightingaleTrackCanvas

midlik:
This PR adds a new component `nightingale-track-canvas` - an alternative to `nightingale-track`, using HTML canvas for rendering instead of SVG graphics.


packages/nightingale-track-canvas/src/utils/utils.ts
```diff
@@ -0,0 +1,208 @@
...
+     * Instead sorts the ranges by their start (ranges with the same start are sorted by decreasing length). */
+    overlappingItemIndices(start: number, stop: number): number[] {
+        const partialOuts = this.binSpans.map(binSpan => this.overlappingItemIndicesInBin(binSpan, start, stop, this._tmpArrays[binSpan] ??= []));
+        return mergeSortedArrays(partialOuts, this.compareFn);
+    }
+
+    private overlappingItemIndicesInBin(binSpan: number, start: number, stop: number, out: number[]): number[] {
+        out.length = 0;
+        const bin = this.bins[binSpan];
+        const from = firstGteqIndex(bin, Math.floor(start) - binSpan + 1, i => this.starts[i]);
```
dlrice, line 151: Not sure I understand why `Math.floor(start) - binSpan + 1` is done?

packages/nightingale-track-canvas/src/utils/utils.ts
```diff
@@ -0,0 +1,208 @@
...
+    protected readonly items: T[];
+    protected readonly starts: number[];
+    protected readonly stops: number[];
+
+    /** Keys to `this.bins`, sorted in ascending order */
+    protected readonly binSpans: number[];
+    /** `this.bins[span]` contains indices of all items whose length is `<= span` but `> span/Q`, in ascending order */
+    protected readonly bins: Record<number, number[]>;
+    /** Ratio of spans of neighboring bins */
+    protected readonly Q = 2;
```
dlrice, line 104: Curious how 2 was selected?

packages/nightingale-track-canvas/src/utils/utils.ts
```diff
@@ -0,0 +1,208 @@
...
+    // key(sortedArray[i]) >= query for each i >= end
+    while (end - start > 4) {
+        const mid = (start + end) >> 1;
+        if (key(sortedArray[mid]) >= query) {
+            end = mid;
+        } else {
+            start = mid + 1;
+        }
+    }
+    // Linear search remaining 4 or fewer elements:
```
dlrice, line 82: Why not continue with the binary search above?

packages/nightingale-track-canvas/src/utils/utils.ts
```diff
@@ -0,0 +1,208 @@
...
+
+/** Return index of the first element within range [start, end) for which `key(element) >= query`.
+ * Return `end` if for all elements in the range `key(element) < query`. */
+function firstGteqIndexInRange<T>(sortedArray: ArrayLike<T>, query: number, start: number, end: number, key: (element: T) => number) {
+    if (start === end) return start;
+    // Invariants:
+    // key(sortedArray[i]) < query for each i < start
+    // key(sortedArray[i]) >= query for each i >= end
+    while (end - start > 4) {
+        const mid = (start + end) >> 1;
```
dlrice, line 75:
```suggestion
        // Floored mean of start and end
        const mid = (start + end) >> 1;
```

packages/nightingale-track-canvas/src/utils/utils.ts
```diff
@@ -0,0 +1,208 @@
...
+        const { start, stop } = accessors;
+        this.items = items;
+        this.starts = items.map(start);
+        this.stops = items.map(stop);
+
+        this.bins = {};
+        for (let i = 0; i < items.length; i++) {
+            const length = this.stops[i] - this.starts[i];
+            let binSpans = 1;
+            while (binSpans < length) binSpans *= this.Q;
```
dlrice, line 119: If you wanted to avoid the while loop you could do
```
const binSpans = this.Q ** Math.ceil(Math.log(length) / Math.log(this.Q));
```
and even pull out the final log above the for loop eg:
```
const logQ = Math.log(this.Q);
...
const binSpans = this.Q ** Math.ceil(Math.log(length) / logQ);
```

packages/nightingale-track-canvas/src/utils/utils.ts
```diff
@@ -0,0 +1,208 @@
...
+    constructor(items: T[], accessors: { start: (item: T) => number, stop: (item: T) => number }) {
+        const { start, stop } = accessors;
+        this.items = items;
+        this.starts = items.map(start);
+        this.stops = items.map(stop);
+
+        this.bins = {};
+        for (let i = 0; i < items.length; i++) {
+            const length = this.stops[i] - this.starts[i];
+            let binSpans = 1;
```
dlrice, line 118:
```suggestion
            let binSpan = 1;
```
I think this is a single bin span as opposed to the later `this.binSpans`.

packages/nightingale-track-canvas/src/nightingale-track-canvas.ts
```diff
@@ -0,0 +1,316 @@
+import { createEvent, customElementOnce } from "@nightingale-elements/nightingale-new-core";
+import NightingaleTrack, { Feature, FeatureLocation, Shapes } from "@nightingale-elements/nightingale-track";
+import { BaseType, select, Selection } from "d3";
+import { html } from "lit";
+import { drawRange, drawSymbol, drawUnknown, shapeCategory } from "./utils/draw-shapes";
+import { last, RangeCollection, Refresher } from "./utils/utils";
+
+
+type Fragment = FeatureLocation["fragments"][number]
+type ExtendedFragment = Fragment & { featureIndex: number, location: FeatureLocation };
```
dlrice, line 10: Is `location: FeatureLocation` needed?

packages/nightingale-track-canvas/src/utils/utils.ts
midlik: Items in that bin have length <=binSpan, so item_stop <= item_start + binSpan. 
Overlapping items must fulfil: 
    item_stop > start
Which can be transformed to:
    item_start + binSpan > start 
    item_start > start - binSpan
And the items have integer start and end, so:
    item_start >= Math.floor(start) - binSpan + 1
    
But now that you pointed out, the last step is not really necessary and it just complicate things. I will simplify it. Thanks

packages/nightingale-track-canvas/src/utils/utils.ts
midlik: Could be any number >1 in theory. But 2 works nice and gives easy-to-work-with spans (1,2,4,8,16...).
With higher ratios you have weak assumption about item length, so you need to check more items that are in fact not overlapping -> worse performance.
On the other hand, smaller ratios (e.g. 1.5) could produce too many bins, which would be more likely to be empty -> worse performance; plus the spans would be really ugly (1,1.5,2.25, 3.375...).

To be more rigorous, the optimal value Q depends on the expected number of items and sequence length (which we don't know exactly so I picked a nice number :D)

packages/nightingale-track-canvas/src/utils/utils.ts
midlik: Thanks, i'm adding a comment here.

packages/nightingale-track-canvas/src/utils/utils.ts
midlik: true

packages/nightingale-track-canvas/src/utils/utils.ts
midlik: I agree it would be faster, but I wanted to avoid float calculations because of imprecisions.
For some numbers, `Math.ceil(Math.log(length) / Math.log(this.Q))` could give one more or one less than expected. Trying to fix that would include some ifs or whiles anyway.

So I chose code simplicity over performance here. Do you think we should go for the log method?
(Current implementation has complexity log(length) which is still not bad.)

packages/nightingale-track-canvas/src/utils/utils.ts
```diff
@@ -0,0 +1,208 @@
...
+        requested = true;
+        if (!running) {
+            handleRequests(); // do not await
+        }
+    }
+    async function handleRequests(): Promise<void> {
+        while (requested) {
+            requested = false;
+            running = true;
+            await sleep(0); // let other things happen (this pushes the rest of the function to the end of the queue)
```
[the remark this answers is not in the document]
midlik: If the `refresh` function takes a long time (more than screen frame time), and `requestRefresh` is called many times (e.g. when touchpad-zooming), then all `refresh`es will be skipped until the first one has finished.

Can `window.requestAnimationFrame` achieve this?

packages/nightingale-track-canvas/tests/nightingale-track-canvas.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    for (const query of rangeQueries) {
+      const truth = overlappingItems_reference(mockRanges, ...query);
+      const found = collection.overlappingItems(...query);
+      expect(found).toEqual(truth);
+    }
+  });
+});
+
+
+function firstGteqIndex_reference<T>(sortedArray: T[], query: number, key: (element: T) => number): number {
```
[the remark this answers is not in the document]
midlik: I think `key` makes more sense here

packages/nightingale-track-canvas/src/nightingale-track-canvas.ts
midlik: You're right, i don't use it anywhere. I'll remove it.

dlrice (approved):

PR #311, opened 2024-11-07
NightingaleScrollbox

midlik:
This PR adds `nightingale-scrollbox`, a container that allows having many tracks in a scrollable box and initializing/disposing them as they become visible/hidden. This can be used to reduce the actual number of tracks and improves performance.


packages/nightingale-scrollbox/src/nightingale-scrollbox.ts
```diff
@@ -0,0 +1,157 @@
...
+        this.disableScrollWithCtrl(this["disable-scroll-with-ctrl"]);
+      }
+    }
+  }
+
+  private initObserver() {
+    const margin = normalizeCssLength(this["root-margin"]);
+    this.observer = new IntersectionObserver(entries => this.observerCallback(entries), { root: this, rootMargin: `${margin} 0px ${margin} 0px` });
+    const items = Array.from(this.items);
+    for (const item of items) {
```
dlrice, line 59:
```suggestion
    for (const item of this.items) {
```
Curious why you needed the `Array.from`?

dlrice (approved):

packages/nightingale-scrollbox/src/nightingale-scrollbox.ts
midlik: You're right, it's not needed here. I probably copied it from elsewhere. 
(on line 100 it's necessary to have Array.from, otherwise we would be deleting elements from a set while iterating it).

PR #318, opened 2025-02-28
Tag shapes

matthiasblum:
This PR adds two new shapes for `nightingale-track`: `tag` and `doubleEndedTag`.

### DIFF
--- a/packages/nightingale-track/src/FeatureShape.ts
+++ b/packages/nightingale-track/src/FeatureShape.ts
@@ -313,7 +313,7 @@ export default class FeatureShape {
   tag() {
     const h = this.#ftHeight;
     const w = this.#ftWidth;
-    const d1 = Math.sqrt(2 * Math.pow(h, 2));
+    const d1 = Math.sqrt(2) * h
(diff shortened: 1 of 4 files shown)


packages/nightingale-track/src/FeatureShape.ts
```diff
@@ -304,4 +309,35 @@ export default class FeatureShape {
     }Z`;
     return rect + triangle;
   }
+
+  tag() {
+    const h = this.#ftHeight;
+    const w = this.#ftWidth;
+    const d1 = Math.sqrt(2 * Math.pow(h, 2));
```
dlrice, line 316:
```suggestion
    const d1 = Math.sqrt(2) * h
```
This seems equivalent, unless we're guarding specifically against `h` < 0?

dlrice (commented):
Would you mind updating `/packages/nightingale-track/tests/mockData/data.json` so the story shows these new shapes too?

matthiasblum:
- rewrote the method to generate the `tag` shape to support three shapes: `leftEndedTag`, `rightEndedTag`, `doubleEndedTag`
- mock data updated
- fixed typos in the README

dlrice (approved):

PR #319, opened 2025-03-05
Adaptation of canvas-track

Swaathik:
Include logic to highlight residues present in a feature for Canvas track which was present in Nightingale-track. 
Convert feature.start and feature.end to Number while calculating fragmentLength as it was giving long numbers from String concatenation sometimes.

### DIFF
--- a/packages/nightingale-track/src/config.ts
+++ b/packages/nightingale-track/src/config.ts
@@ -305,8 +305,8 @@ export const config = {
     shape: "rectangle",
     color: "#8585fc",
   },
-  PTM: {
-    name: "ptm",
+  RESIDUE: {
+    name: "residue",
     label: "PTM residue",
     tooltip: "",
     shape: "rectangle",
--- a/packages/nightingale-track/src/nightingale-track.ts
+++ b/packages/nightingale-track/src/nightingale-track.ts
@@ -32,7 +32,7 @@ export type FeatureLocation = {
 };
 
 type Residue = {
-  name: string;
+  name?: string;
   position: number;
 };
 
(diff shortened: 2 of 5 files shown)


packages/nightingale-track/src/nightingale-track.ts
```diff
@@ -203,16 +183,18 @@ class NightingaleTrack extends withManager(
...
-    // For longer proteins, the PTMs have to be shown prominent in the first look
-    if (this.length && this.length > 500) {
-      ptmLength = this.getSingleBaseWidth() < 4 ? 4 : 1;
-    }
+   #getResidueShape(f: Residue & { feature: Feature; position: number }) {
+    let residueLength = 1;
+     // Below logic is to show it prominent for longer proteins until the point where residueLength is enough to be visible on itself.
+     const optimalWidth = 6;
+     const widthDifference = optimalWidth - this.getSingleBaseWidth();
+     if (this.getSingleBaseWidth() < optimalWidth && widthDifference > residueLength) {
```
dlrice, line 191: Could just assign to a variable eg `const singleBaseWidth = this.getSingleBaseWidth()` rather than calling twice.

packages/nightingale-track/src/config.ts
```diff
@@ -305,8 +305,8 @@ export const config = {
     shape: "rectangle",
     color: "#8585fc",
   },
-  PTM: {
-    name: "ptm",
+  RESIDUE: {
+    name: "residue",
     label: "PTM residue",
```
dlrice, line 310: Does this need to change too?

Swaathik: https://github.com/ebi-webcomponents/nightingale/blob/canvas-uniprot-fix/packages/nightingale-track-canvas/src/nightingale-track-canvas.ts#L158, I call this in two places (nightingale-track and canvas) where 'PTM' is out of context.

dlrice (approved):

PR #326, opened 2025-10-13
Change AlphaFoldPayload variable name and update type.

dlrice:


packages/nightingale-structure/src/nightingale-structure.ts
```diff
@@ -48,26 +48,50 @@ export type PDBData = Record<
   }
 >;
 
-export type PredictionData = {
+export type AlphaFoldPayload = Array<{
   entryId: string;
+  modelEntityId?: string;
```
Swaathik: EntryId is deprecated and renamed to `modelEntityId` in the recent AlphaFold update. Shall we replace the few occurrences of entryId with modelEntityId as we need to do it at some point anyway?

Swaathik:
```suggestion
  modelEntityId: string;
```

packages/nightingale-structure/src/nightingale-structure.ts
```diff
@@ -48,26 +48,50 @@ export type PDBData = Record<
...
-  distogramUrl?: string;
+  paeImageUrl?: string;
+  msaUrl?: string;
+  plddtDocUrl?: string;
+  paeDocUrl?: string;
+  uniprotSequence?: string;
+  uniprotStart?: number;
+  uniprotEnd?: number;
+  isReferenceProteome?: boolean;
+  isReviewed?: boolean;
```
Swaathik: Also, these ones are deprecated and the new fields are included. Safe to remove these too as they are only part of the schema and not used anywhere else

Swaathik (changes requested):

packages/nightingale-structure/src/nightingale-structure.ts
dlrice: Good point - updated

Swaathik (approved):

PR #327, opened 2026-03-26
Security audit

dlrice:
- Fix XSS in nightingale-scrollbox: Sanitize innerHTML assignments in setContent() using DOMPurify, preventing script injection via content-visible/content-hidden attributes
- Fix XSS in nightingale-structure: Sanitize the .innerHTML Lit property binding in the message overlay using DOMPurify
- Remove typosquatted rimraff dependency: Delete the orphaned rimraff@0.0.1-security entry from yarn.lock (no package.json referenced it)
- Add security documentation to inline worker pattern: Document constraints on the code-as-string Web Worker construction in nightingale-msa to prevent future misuse
- Add dompurify as runtime dependency to nightingale-scrollbox and nightingale-structure package.json files
- Update github action node version to 24

### DIFF
--- a/.github/workflows/test-and-publish-app.yml
+++ b/.github/workflows/test-and-publish-app.yml
@@ -9,7 +9,7 @@ jobs:
       - uses: actions/checkout@v6.0.2
       - uses: actions/setup-node@v6.3.0
         with:
-          node-version: 20
+          node-version: 24
       - name: 🔧 Install
         run: |
           yarn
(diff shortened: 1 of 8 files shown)


.github/workflows/test-and-publish-app.yml
```diff
@@ -6,8 +6,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: actions/checkout@v6.0.2
+      - uses: actions/setup-node@v6.3.0
         with:
           node-version: 20
```
Swaathik: They are deprecating version 20 and recommended node 24 for github runner. We can also update it as part of the change

Swaathik (approved):

PR #328, opened 2026-04-15
Added LOGO track

pmustonebi:


packages/nightingale-logo-track/src/nightingale-logo-track.ts
```diff
@@ -0,0 +1,188 @@
...
+    const marginRight = this["margin-right"];
+    const marginColor = this["margin-color"];
+    const trackHeight = height - marginTop - marginBottom;
+
+    if (trackHeight <= 0) return;
+
+    const start = Math.round(this["display-start"] ?? 1);
+    const end = Math.round(this["display-end"] ?? this.length ?? start);
+
+    if (end < start) return;
```
dlrice: When the user doesn't set `display-end` the position mixin defaults it to a special value, `-1`, meaning "show everything to the end" (it's named `WHOLE_SEQ` in `nightingale-new-core`).

Because `??` only falls back on `null`/`undefined` (not `-1`), `end` becomes `-1`, and the `if (end < start) return;` on the next line bails out so nothing is drawn.  `withZoom` already handles this, we just need to do the same when choosing `end`:

```ts
const rawEnd = this["display-end"];
const end = Math.round(rawEnd && rawEnd !== -1 ? rawEnd : (this.length ?? start));
```

packages/nightingale-logo-track/src/nightingale-logo-track.ts
```diff
@@ -0,0 +1,188 @@
...
+    const letters: LetterData[] = [];
+    for (const nuc of TRACKED_NUCLEOTIDES) {
+      const f = counts[nuc] / total;
+      if (f > 0) {
+        letters.push({ nucleotide: nuc, heightFraction: (f * IC) / LOG2_4 });
+      }
+    }
+
+    // Most-conserved nucleotide at the bottom of the stack
+    letters.sort((a, b) => b.heightFraction - a.heightFraction);
```
dlrice: I noticed on other logo plots eg https://weblogo.threeplusone.com/examples.html they have the largest letters on top while this PR has them on the bottom. I just wanted to doublecheck which way you want it?

packages/nightingale-logo-track/src/nightingale-logo-track.ts
```diff
@@ -0,0 +1,188 @@
...
+}
+
+export interface MsaSequence {
+  name: string;
+  sequence: string;
+}
+
+@customElementOnce("nightingale-logo-track")
+export default class NightingaleLogoTrack extends withManager(
+  withZoom(withHighlight(NightingaleElement))
```
dlrice: The component mixes in `withHighlight` currently doesn't draw the highlighted region or reacts to those attributes. `withHighlight` on its own only tracks which region is highlighted and then each component then needs to implement how the highlight is rendered.

There's a ready-made mixin for exactly this: `withSVGHighlight` (see how `nightingale-variation` uses it). It gives you `createHighlightGroup()` and a working `updateHighlight()`. It would involve calling `createHighlightGroup()` in `createTrack()` and `updateHighlight()` inside `refresh()`, and the highlight rectangle should then work.

packages/nightingale-logo-track/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "dist",
+    "src"
+  ],
+  "main": "dist/index.js",
+  "module": "dist/index.js",
+  "type": "module",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "build": "rollup --config ../../rollup.config.mjs",
+    "test": "../../node_modules/.bin/jest --config ../../jest.config.js ./tests/*"
```
dlrice, line 15: The `test` script runs jest against `./tests/*`, but there's no `tests` folder in the package yet, so `yarn test` here doesn't actually run anything.

packages/nightingale-logo-track/src/nightingale-logo-track.ts
```diff
@@ -0,0 +1,188 @@
...
+      }
+    }
+
+    // Draw margin overlays to mask letters that extend into margin areas
+    parts.push(
+      `<rect x="0" y="0" width="${marginLeft}" height="${height}" fill="${marginColor}" pointer-events="none"/>`,
+      `<rect x="${width - marginRight}" y="0" width="${marginRight}" height="${height}" fill="${marginColor}" pointer-events="none"/>`,
+      `<rect x="${marginLeft}" y="0" width="${width - marginLeft - marginRight}" height="${marginTop}" fill="${marginColor}" pointer-events="none"/>`,
+      `<rect x="${marginLeft}" y="${height - marginBottom}" width="${width - marginLeft - marginRight}" height="${marginBottom}" fill="${marginColor}" pointer-events="none"/>`
+    );
```
dlrice: This assembles the SVG as one big string and assigns it with `innerHTML` on every refresh. It works, but two things:

1. `refresh()` runs on every zoom/pan frame, and this throws away and re-parses the whole column DOM each time which takes bit of time.
2. The other tracks (have a look at `nightingale-linegraph-track`) build and update their SVG with d3's "data join" — `selectAll(...).data(...).join("text")` which only touches the elements that actually changed and keeps the code consistent across the repo.

Moving to a d3 join would make it both faster and more in keeping with the rest of the codebase.

packages/nightingale-logo-track/src/nightingale-logo-track.ts
```diff
@@ -0,0 +1,188 @@
...
+            `text-anchor="middle" ` +
+            `font-family="Arial,Helvetica,sans-serif" font-size="${FONT_SIZE}" font-weight="bold" ` +
+            `fill="${color}">${item.nucleotide}</text>`
+        );
+      }
+    }
+
+    // Draw margin overlays to mask letters that extend into margin areas
+    parts.push(
+      `<rect x="0" y="0" width="${marginLeft}" height="${height}" fill="${marginColor}" pointer-events="none"/>`,
```
dlrice: The `withMargin` mixin you're already using provides a `renderMarginOnGroup(group)` helper that draws exactly the four rectangles margin for you (and reuses them rather than recreating them). `nightingale-linegraph-track` shows the usage. Switching to it is less code to maintain, and you'll stay in sync automatically if the shared margin behaviour ever changes.

packages/nightingale-logo-track/src/nightingale-logo-track.ts
```diff
@@ -0,0 +1,188 @@
...
+  withZoom,
+} from "@nightingale-elements/nightingale-new-core";
+import { select } from "d3";
+import { html, PropertyValues } from "lit";
+
+const NUCLEOTIDE_COLORS: Record<string, string> = {
+  A: "#00CC00",
+  C: "#0000CC",
+  G: "#FFB300",
+  U: "#CC0000",
```
dlrice, line 13: Would you consider extending to work with amino acids too?

pmustonebi: Sure! Updated to add this capability

packages/nightingale-logo-track/src/nightingale-logo-track.ts
pmustonebi: updated to set the largest letter at the top.

packages/nightingale-logo-track/src/nightingale-logo-track.ts
pmustonebi: Should be resolved, please check

packages/nightingale-logo-track/package.json
pmustonebi: Should be resolved, please check

packages/nightingale-logo-track/src/nightingale-logo-track.ts
pmustonebi: Should be resolved, please check

packages/nightingale-logo-track/src/nightingale-logo-track.ts
pmustonebi: Should be resolved, please check

packages/nightingale-manager/src/nightingale-manager.ts
```diff
@@ -93,31 +93,33 @@ class NightingaleManager extends NightingaleElement {
   register(element: NightingaleElement) {
     this.htmlElements.add(element);
     this.applyAttributesOnElement(element);
+    this.applyPropertiesOnElement(element);
```
dlrice: Changing `register()` to `applyPropertiesOnElement` bypasses the `for` target restriction. The manager now puts the entire `propertyValues` map onto every element during registration. This causes track-specific data (like a `filters` value intended solely for Track A) to leak onto any track that connects later, quietly breaking `nightingale-filter`'s targeting.

I would suggest to:
* Revert the changes made to `nightingale-manager.ts`.
* In the Storybook files, set `logoTrack.sequences = [...]` directly on the element instead of dispatching a manager `change` event. (This matches how the README's standalone example handles it, e.g., `(linegraphTrack as any).data = ...`).
* Note that global view-state (`length`, `display-start/-end`, `highlight`) already flows correctly via attributes and mixins.

packages/nightingale-logo-track/src/nightingale-logo-track.ts
```diff
@@ -0,0 +1,251 @@
...
+  // Positively charged
+  K: "#0000CC", R: "#0000CC", H: "#0000CC",
+  // Negatively charged
+  D: "#CC0000", E: "#CC0000",
+};
+
+const TRACKED_NUCLEOTIDES = ["A", "C", "G", "U"];
+const TRACKED_AMINO_ACIDS = ["A", "C", "D", "E", "F", "G", "H", "I", "K", "L", "M", "N", "P", "Q", "R", "S", "T", "V", "W", "Y"];
+/** Characters that appear in protein sequences but not in RNA/DNA */
+const AMINO_ACID_ONLY_CHARS = new Set(["D", "E", "F", "H", "I", "K", "L", "M", "N", "P", "Q", "R", "S", "V", "W", "Y"]);
```
dlrice: `AMINO_ACID_ONLY_CHARS` currently includes the character `N`. However, `N` is the standard IUPAC code for an ambiguous/unknown nucleotide. If a nucleotide alignment contains more than 5% `N`s, `detectSeqType` (line 107) misclassifies the sequence as a protein. This also  contradicts line 103 of the README, which explicitly lists `N` as a nucleotide.

I suggest to remove `N` from `AMINO_ACID_ONLY_CHARS`.

packages/nightingale-logo-track/package.json
dlrice: I still can't see any tests within the branch. Maybe these were accidentally omitted? You can see the test conventions in other packages e.g.:

```
$ git:(add-logo-track) ✗ find packages -name "tests" -type d
packages/nightingale-structure/src/tests
packages/nightingale-navigation/dist/nightingale-filter/tests
packages/nightingale-navigation/tests
packages/nightingale-sequence/tests
packages/nightingale-track/tests
packages/nightingale-linegraph-track/tests
packages/nightingale-new-core/tests
packages/nightingale-sequence-heatmap/tests
packages/nightingale-manager/dist/nightingale-filter/tests
packages/nightingale-manager/tests
packages/nightingale-track-canvas/tests
packages/nightingale-variation/tests
packages/nightingale-conservation-track/tests
packages/nightingale-interpro-track/tests
packages/nightingale-saver/dist/nightingale-filter/tests
packages/nightingale-filter/tests
packages/nightingale-textarea-sequence/tests
packages/nightingale-colored-sequence/tests
packages/nightingale-links/tests
```

packages/nightingale-logo-track/src/nightingale-logo-track.ts
pmustonebi: Should be resolved, please check

packages/nightingale-logo-track/package.json
pmustonebi: Should be resolved, please check

packages/nightingale-manager/src/nightingale-manager.ts
pmustonebi: Should be resolved, please check

packages/nightingale-logo-track/src/nightingale-logo-track.ts
pmustonebi: Resolved, please check :)

PR #331, opened 2026-06-05
nightingale-distribution-track

midlik:
This PR adds `nightingale-boxplot-track` for showing distribution of a variable on each residue position via boxplots.

Screenshot-nightingale-boxplot-track


packages/nightingale-boxplot-track/src/nightingale-boxplot-track.ts
```diff
@@ -0,0 +1,983 @@
...
+import { Downsampler } from "./downsampling";
+
+
+const ATTRIBUTES_THAT_TRIGGER_REFRESH: string[] = [
+  "length", "width", "height",
+  "margin-top", "margin-bottom", "margin-left", "margin-right", "margin-color",
+  "y-min", "y-max", "show-axis", "zoomed-out-outline",
+  "column-gap", "box-gap", "whisker-width", "outlier-jitter-width", "outlier-radius", "zoom-transition-range",
+] satisfies (keyof NightingaleBoxplotTrack)[];
+const ATTRIBUTES_THAT_TRIGGER_DATA_RESET: string[] = [] satisfies (keyof NightingaleBoxplotTrack)[];
```
Swaathik: it is not populated anywhere else but being checked if any value is present? Is it needed?

packages/nightingale-new-core/src/utils/attribute-converters.ts
```diff
@@ -0,0 +1,65 @@
...
+ *   "y-min"?: number;
+ * }
+ * ```
+ * */
+export const OptionalNumberAttributeConverter: ComplexAttributeConverter<number | undefined> = {
+    fromAttribute(str) {
+        if (!str) { // null or empty string
+            return undefined;
+        } else {
+            return Number(str);
```
Swaathik: should there be an additional check for NaN 
`isNaN(Number(str))` here as well before returning?

midlik: ✔ I added `isNan` check

packages/nightingale-boxplot-track/src/nightingale-boxplot-track.ts
midlik: You're right, this was copied from `NightingaleConservationTrack` and the list became empty.

✔ I removed it.

Swaathik (approved):

